PR #13, opened 2021-02-17
Add RunnableTask type and custom properties

regadas:
The main idea here is to allow setting/override `type` and `custom` when defining other types of `RunnableTask`'s.

### DIFF
--- a/flytekit-api/src/main/java/org/flyte/api/v1/RunnableTask.java
+++ b/flytekit-api/src/main/java/org/flyte/api/v1/RunnableTask.java
@@ -17,6 +17,7 @@
 package org.flyte.api.v1;
 
 import java.util.Map;
+import com.google.protobuf.Struct;
 
 /** Building block for tasks that execute Java code. */
 public interface RunnableTask {
@@ -28,4 +29,8 @@ public interface RunnableTask {
   Map<String, Literal> run(Map<String, Literal> inputs);
 
   RetryStrategy getRetries();
+
+  String getType();
+
+  Struct getCustom();
 }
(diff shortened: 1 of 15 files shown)


flytekit-api/src/main/java/org/flyte/api/v1/RunnableTask.java
```diff
@@ -28,4 +29,8 @@
   Map<String, Literal> run(Map<String, Literal> inputs);
 
   RetryStrategy getRetries();
+
+  String getType();
+
+  Struct getCustom();
```
regadas, line 35: I reckon that having type `Struct` is probably not the desirable thing given the design of the rest of the API. I believe we could represent it as `Map<String, Literal>`, would that be better?

pom.xml
```diff
@@ -164,6 +164,11 @@
         <artifactId>protobuf-java</artifactId>
         <version>${protobuf.version}</version>
       </dependency>
+      <dependency>
+        <groupId>com.google.protobuf</groupId>
+        <artifactId>protobuf-java-util</artifactId>
```
regadas, line 169: Gonna rm this as it's not needed yet.

flytekit-api/src/main/java/org/flyte/api/v1/RunnableTask.java
regadas: This commit [`8649252` (#13)](https://github.com/spotify/flytekit-java/pull/13/commits/86492525292106d5eeb539036eae8d4a50bef882) addresses the above lml wdyt

regadas:
Talked IRL @kanterov closing this.

PR #22, opened 2021-06-04
Added github workflow

evalsocket:
- Added GitHub workflow for the pull request and release snapshot for each master merge. 

- [ ] Codecovare is not implimented 

Fixes : 
https://github.com/flyteorg/flyte/issues/995
https://github.com/flyteorg/flyte/issues/996
https://github.com/flyteorg/flyte/issues/994


evalsocket:
@EngHabu Is there a way to verify the workflow before the merge? Release workflow is not tested.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,84 @@
...
+      uses: sonatype-nexus-community/nexus-repo-github-action@master
+      with:
+        serverUrl: ${{ secrets.SONATYPE_SERVER_URL }}
+        username: ${{ secrets.SONATYPE_USERNAME }}
+        password: ${{ secrets.SONATYPE_PASSWORD }}
+        format: maven2
+        repository: flytekit-api
+        coordinates: groupId=org.flyte artifactId=flytekit-parent version=${{ steps.bump.outputs.version }}
+        assets: extension=jar
+        filename: ./flytekit-api/target/flytekit-api-${{ steps.bump.outputs.version }}.jar
```
kanterov, line 38: I'm not sure it's going to work. Maven is going to use a version from the pom file. It will be great if we followed a similar process as for instance in Netty. Here is their CI release setup:

https://github.com/netty/netty/blob/4.1/.github/workflows/ci-release.yml

In this case a bot will bump the version:

https://github.com/netty/netty/commit/9e06241ddbe4d3aedc7994d37450013de276f062

With that, we also don't need custom GitHub actions and can use Maven.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,84 @@
+name: Flytekit Java Release
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
```
kanterov, line 9: Should we add a condition to make sure it doesn't run outside of main repo? E.g.
```
if: github.repository == 'flyteorg/flytekit-java'
```

kanterov (changes requested):
I think we need to iterate on release part. We can look at how it's done in other OSS projects using Maven, and follow the same approach. A good example can be Netty. Except for this, I think we miss a couple more pieces:
- testing for pull requests (only main branch has action)
- publishing of snapshots jars and Docker images (should be done for each build of the main branch)

A good start can be handling the release of jars and Docker images in a separate pull request. And limit the scope of the initial pull request to running `mvn verify` on PRs and main branch builds.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,38 @@
+name: Flytekit Java Build
```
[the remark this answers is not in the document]
evalsocket: ok

.github/workflows/build.yaml
```diff
@@ -0,0 +1,38 @@
...
+      with:
+        fetch-depth: "0"
+    - name: Set up JDK 1.8
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+    - name: Verify with Maven
+      run: mvn verify
+    - name: Build with Maven
+      run: mvn package
```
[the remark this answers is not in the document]
evalsocket: ok

.github/workflows/release.yaml
evalsocket: No, it is not required for release because we are using a release trigger.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,48 @@
...
+#              "username": "${{ secrets.SONATYPE_USERNAME }}",
+#              "password": "${{ secrets.SONATYPE_PASSWORD }}"
+#            }]
+      - name: Verify with Maven
+        run: mvn verify
+      - name: Release snapshot with Maven
+        if: ${{ github.ref == 'refs/heads/master' }}
+        env:
+          RELEASE_REPOSITORY_URL: "https://s01.oss.sonatype.org/"
+          SNAPSHOTS_REPOSITORY_URL: "https://s01.oss.sonatype.org/"
```
kanterov, line 47: I think it should be `https://s01.oss.sonatype.org/content/repositories/snapshots`

.github/workflows/build.yaml
```diff
@@ -0,0 +1,48 @@
...
+#              "id": "flytekit-java-release",
+#              "username": "${{ secrets.SONATYPE_USERNAME }}",
+#              "password": "${{ secrets.SONATYPE_PASSWORD }}"
+#            }]
+      - name: Verify with Maven
+        run: mvn verify
+      - name: Release snapshot with Maven
+        if: ${{ github.ref == 'refs/heads/master' }}
+        env:
+          RELEASE_REPOSITORY_URL: "https://s01.oss.sonatype.org/"
```
kanterov, line 46: I think it should be `https://s01.oss.sonatype.org/service/local/staging/deploy/maven2/`

.github/workflows/build.yaml
```diff
@@ -0,0 +1,48 @@
...
+        uses: actions/cache@v2
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-prepare-release-${{ hashFiles('pom.xml') }}
+      - name: Set up JDK 11
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11.0
+          distribution: 'adopt'
+#      - name: Log in to Docker Hub
```
kanterov, line 27: Why is this commented out?

.github/workflows/build.yaml
```diff
@@ -0,0 +1,48 @@
...
+#              "password": "${{ secrets.SONATYPE_PASSWORD }}"
+#            }]
+      - name: Verify with Maven
+        run: mvn verify
+      - name: Release snapshot with Maven
+        if: ${{ github.ref == 'refs/heads/master' }}
+        env:
+          RELEASE_REPOSITORY_URL: "https://s01.oss.sonatype.org/"
+          SNAPSHOTS_REPOSITORY_URL: "https://s01.oss.sonatype.org/"
+        run: mvn deploy -DpreparationGoals=clean -DskipTests=true -Dghcr.io/flyteorg/jflyte  -Ddockerfile.push
```
kanterov, line 48: Should it be `-Ddocker.image=gcr.io/flyteorg/jflyte`? Probably we can name image as flytekit-java to make it less confusing as well.

kanterov (changes requested):
I'm not sure 'flytekit-java release" works as intended. In one of the comments before, I mentioned that we should do something close to how Netty does releases. In Netty, the GitHub bot will push commit to change the version in pom.xml. And then perform release against this commit. This is how maven-release-plugin works for instance.

This takes more time to implement. What if we start with publishing only snapshots, and remove this code for now?

.github/workflows/build.yaml
evalsocket: It's failing...I think we don't have secrets in place for docker. Can you check ?

evalsocket:
I removed the release workflow. And only enable build and push workflow for master

kanterov:
@evalsocket should "Log in to Docker Hub" be done only for the main branch build, similar to how "Release snapshot with Maven" works?

kanterov (approved):
Nice! Thanks

PR #28, opened 2021-07-16
New copyright headers will be copyrighted to Flyte Authors.

kumare3:
Signed-off-by: Ketan Umare <ketan.umare@gmail.com>

# TL;DR
 - Standardized contributing guidelines
 - Readme
 - Also added additional copyright to Flyte Authors.

## Type
 - [x] Docs only

### DIFF
--- a/spotbugs-exclude.xml
+++ b/spotbugs-exclude.xml
@@ -1,6 +1,5 @@
 <?xml version="1.0"?>
 <!--
-  Copyright 2020 Spotify AB.
   Copyright 2021 Flyte Authors.
 
   Licensed under the Apache License, Version 2.0 (the "License");
--- a/spotbugs-exclude.xml
+++ b/spotbugs-exclude.xml
@@ -1,6 +1,7 @@
 <?xml version="1.0"?>
 <!--
   Copyright 2020 Spotify AB.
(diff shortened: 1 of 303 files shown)


kanterov:
@kumare3 can you do `mvn spotless:apply`?

kumare3:
> @kumare3 can you do `mvn spotless:apply`?

Will do

ckiosidis (approved):

suppressions.xml
```diff
@@ -1,6 +1,7 @@
 <?xml version="1.0"?>
 <!--
   Copyright 2020 Spotify AB.
```
lupasarin, line 3: Shouldn't this line be removed here too?

spotbugs-exclude.xml
```diff
@@ -1,6 +1,7 @@
 <?xml version="1.0"?>
 <!--
   Copyright 2020 Spotify AB.
```
lupasarin, line 3: Shouldn't this line be removed here too?

lupasarin (approved):

suppressions.xml
narape: Right, doing a find in the whole project showed more places. I will check further

narape (approved):

PR #50, opened 2021-12-17
SdkRemoteLaunchPlans

ckiosidis:
# TL;DR

Adding support for SdkRemoteLaunchPlan, similarly to the existing SdkRemoteTask.
I will put together a separate PR for the SdkRemoteWorkflow 

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
Created a SdkRemoteLaunchPlan class that implements the apply method to return a SdkNode.
The SdkWorkflowNode/WorkflowNode classes were already implemented previously.

## Tracking Issue
adds to  https://github.com/flyteorg/flyte/issues/1002

## Follow-up issue
_NA_

### DIFF
--- a/flytekit-java/src/main/java/org/flyte/flytekit/SdkRemoteLaunchPlan.java
+++ b/flytekit-java/src/main/java/org/flyte/flytekit/SdkRemoteLaunchPlan.java
@@ -1,0 +1,123 @@
+/*
+ * Copyright 2021 Flyte Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.flyte.flytekit;
+
+import static org.flyte.flytekit.MoreCollectors.toUnmodifiableMap;
+
+import com.google.auto.value.AutoValue;
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Nullable;
+import org.flyte.api.v1.PartialLaunchPlanIdentifier;
+import org.flyte.api.v1.WorkflowNode;
+
+/** Reference to a LaunchPlan deployed in flyte, a remote LaunchPlan. */
+@AutoValue
+public abstract class SdkRemoteLaunchPlan<InputT, OutputT> extends SdkTransform {
+
+  @Nullable
+  public abstract String domain();
+
+  public abstract String project();
+
+  public abstract String name();
+
+  @Nullable
+  public String version() {
+    return null;
+  }
+
+  public abstract SdkType<InputT> inputs();
+
+  public abstract SdkType<OutputT> outputs();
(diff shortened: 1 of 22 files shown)


flytekit-java/src/main/java/org/flyte/flytekit/SdkRemoteLaunchPlan.java
```diff
@@ -0,0 +1,123 @@
...
+
+  @Nullable
+  public abstract String domain();
+
+  public abstract String project();
+
+  public abstract String name();
+
+  @Nullable
+  public String version() {
```
tnsetting, line 40: How would version work?

ckiosidis: When the version is empty, the sdk fetches it from the platform, see https://github.com/flyteorg/flytekit-java/blob/master/jflyte/src/main/java/org/flyte/jflyte/IdentifierRewrite.java#L140-L161

tnsetting: Can we add a test for it somehow?

ckiosidis: The IdentifierRewrite is tested separately, and I added the example here that works when I tested it https://github.com/flyteorg/flytekit-java/blob/b8b5e0e677be7da6d123b81c511e9283790b1484/flytekit-examples/src/main/java/org/flyte/examples/RemoteLaunchPlanExample.java

flytekit-java/src/test/java/org/flyte/flytekit/SdkRemoteLaunchPlanTest.java
```diff
@@ -0,0 +1,135 @@
...
+                        .id("some-node-id")
+                        .workflowNode(
+                            WorkflowNode.builder()
+                                .reference(
+                                    WorkflowNode.Reference.ofLaunchPlanRef(
+                                        PartialLaunchPlanIdentifier.builder()
+                                            .domain("dev")
+                                            .project("project-a")
+                                            .name("SomeWorkflow")
+                                            .version("version")
```
tnsetting, line 72: should be the `version` be null?

flytekit-java/src/main/java/org/flyte/flytekit/SdkRemoteLaunchPlan.java
```diff
@@ -0,0 +1,123 @@
...
+            .domain(domain())
+            .version(version())
+            .build();
+    List<CompilerError> errors = Compiler.validateApply(nodeId, inputs, inputs().getVariableMap());
+
+    if (!errors.isEmpty()) {
+      throw new CompilerException(errors);
+    }
+
+    Map<String, SdkBindingData> outputs =
```
tnsetting, line 83: So we don't validate the input/output of the remote Launch plan.

ckiosidis: We don't double check the remote outputs but the register step still fails if the interface is difference.
This part I copied from another place to keep it consistent

tnsetting: Ok. Cool.

tnsetting (approved):

EngHabu (approved):
With my limited knowledge in this SDK, it looks good to me!

PR #64, opened 2022-02-08
[test deploy] Copy dbeam config

veggiemonk:
Signed-off-by: Julien Bisconti <julienb@spotify.com>

# TL;DR
Test GPG with maven deploy

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -234,6 +234,7 @@
       <distribution>repo</distribution>
     </license>
   </licenses>
+
   <scm>
     <connection>scm:git:https://github.com/flyteorg/flytekit-java.git</connection>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -234,28 +234,65 @@
...
       <organization>Flyte</organization>
       <organizationUrl>https://flyte.org</organizationUrl>
     </developer>
+    <developer>
+      <id>dataex</id>
+      <name>Spotify DataEx</name>
+      <email>dataex@spotify.com</email>
+      <organization>Spotify AB</organization>
+      <organizationUrl>http://www.spotify.com</organizationUrl>
+    </developer>
```
narape, line 259: @veggiemonk I think that we should remove this part. At least we should discuss it in a different PR

PR #72, opened 2022-02-22
[minor] Add sub-workflow support in local execution

RustedBones:
Signed-off-by: Michel Davit <michel@davit.fr>

# TL;DR
Add sub-workflow support in local execution

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
Add the possibility to run and test locally nested workflows. (`SubWorkflowRef` only at the moment)

`WorkflowNode` with `SubWorkflowRef` reference will be expanded.

Nested ids will be get the sub-workflow root node as prefix to allow several usage of the workflow. I tried to follow the [doc statement](https://docs.flyte.org/projects/cookbook/en/latest/auto/core/control_flow/subworkflows.html#example-1)

> Flyte will automatically prepend something to the inner n0, since node IDs need to be distinct within a workflow graph
 
Sub-workflow input are remapped to the parent node input.

## Tracking Issue

https://github.com/flyteorg/flyte/issues/1002


flytekit-java/src/main/java/org/flyte/flytekit/SdkWorkflowBuilder.java
```diff
@@ -33,13 +33,20 @@
...
   private final Map<String, SdkBindingData> outputs;
   private final Map<String, String> inputDescriptions;
   private final Map<String, String> outputDescriptions;
 
-  protected SdkWorkflowBuilder() {
+  public SdkWorkflowBuilder() {
+    this("");
+  }
+
+  public SdkWorkflowBuilder(String idPrefix) {
```
RustedBones, line 47: Visibility change is not desired bur required to expand the workflow in the `SdkTestingExecutor`
We could have another test builder that have a public constructor to keep this one protected.

flytekit-local-engine/src/main/java/org/flyte/localengine/LocalEngine.java
```diff
@@ -50,42 +51,58 @@
   public static Map<String, Literal> compileAndExecute(
       WorkflowTemplate template,
       Map<String, RunnableTask> runnableTasks,
+      Map<String, DynamicWorkflowTask> dynamicWorkflowTasks,
```
RustedBones, line 54: This is a 'shift' on the public API:
I deleted the deprecated API, deprecate the current one and add a new one with `workflows` parameter

flytekit-java/src/main/java/org/flyte/flytekit/SdkWorkflow.java
```diff
@@ -57,13 +58,15 @@ public SdkNode apply(
             .reference(WorkflowNode.Reference.ofSubWorkflowRef(workflowId))
             .build();
 
+    Map<String, SdkBindingData> outputs =
+        innerBuilder.getOutputs().entrySet().stream()
+            .collect(
+                Collectors.toMap(
+                    Map.Entry::getKey,
+                    e ->
+                        SdkBindingData.ofOutputReference(nodeId, e.getKey(), e.getValue().type())));
```
RustedBones, line 67: We need to hide the sub-workflow internal nodes from the output.
Simply provide a out reference binding on the node itself

kumare3 (approved):
This looks really cool @RustedBones. Thank you so much. But, for syntactic understand of flytekit java, maybe @narape / @ckiosidis are better proxies?

RustedBones:
Any feedback on that ?

RustedBones:
https://github.com/RustedBones/flytekit-java/pull/1 would be a follow-up to be able to mock sub workflows in the test suite

RustedBones:
Up :arrow_up:

flytekit-local-engine/src/main/java/org/flyte/localengine/ExecutionNode.java
```diff
@@ -28,10 +30,15 @@ public abstract class ExecutionNode {
 
   public abstract List<String> upstreamNodeIds();
 
+  @Nullable
```
sonjaer, line 33: Does this need to be nullable?

flytekit-local-engine/src/main/java/org/flyte/localengine/LocalEngine.java
```diff
@@ -106,7 +133,20 @@ static Map<String, Literal> execute(
 
       listener.starting(executionNode, inputs);
 
-      Map<String, Literal> outputs = runWithRetries(executionNode, inputs, listener);
+      Map<String, Literal> outputs;
+      if (executionNode.subWorkflow() != null) {
```
sonjaer, line 137: Maybe we could add a test to LocalEngineTest which includes a nested subworkflow to validate that things are executed in the correct order, wdyt?

flytekit-java/src/main/java/org/flyte/flytekit/SdkWorkflow.java
sonjaer: I am a bit unsure what the implication of this is as I find it hard to follow in the code and I am not an expert on this repo. From my understanding this won't impact the protos deployed to Flyte so only Local execution. Would the alternative be to change how we compile the graph in local execution instead to avoid changing the outputs? cc: @rafaelraposospot @narape

sonjaer: I tried the Uber workflow in Flyte, it seems to be working fine

flytekit-local-engine/src/main/java/org/flyte/localengine/LocalEngine.java
rafaelraposospot: Done

flytekit-local-engine/src/main/java/org/flyte/localengine/LocalEngine.java
```diff
@@ -179,6 +220,8 @@ static Literal getLiteral(
 
         requireNonNull(outputs, () -> String.format("missing output for node [%s]", nodeId));
 
+        Literal literalOutputs = outputs.get(bindingData.promise().var());
```
sonjaer, line 223:
```suggestion
```

flytekit-local-engine/src/main/java/org/flyte/localengine/LocalEngine.java
```diff
@@ -179,6 +220,7 @@ static Literal getLiteral(
 
         requireNonNull(outputs, () -> String.format("missing output for node [%s]", nodeId));
 
+
```
sonjaer, line 223:
```suggestion
```

sonjaer (approved):

benkonz:
@RustedBones are there plans to support BranchNode in the test executor?

RustedBones:
That was part of the plan, but I'm not sure I'll get the time to work on it unfortunately

PR #120, opened 2022-07-14
Add support for mocking sub-workflows

RustedBones:
Signed-off-by: Michel Davit <michel@davit.fr>

# TL;DR
Add support for mocking sub-workflows in tests

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Complete description

Give the ability to mock sub-workflow by giving a `Map<String, Literal>` for input and output, in a similar way for Task.
-> As workflows do not have input/output type, generic `Map<String, Literal>` are used
-> Literals constructor need to be made public

### DIFF
--- a/flytekit-examples/src/test/java/org/flyte/examples/WorkflowTest.java
+++ b/flytekit-examples/src/test/java/org/flyte/examples/WorkflowTest.java
@@ -18,6 +18,10 @@ package org.flyte.examples;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
+import java.util.HashMap;
+import java.util.Map;
+import org.flyte.api.v1.Literal;
+import org.flyte.flytekit.testing.Literals;
 import org.flyte.flytekit.testing.SdkTestingExecutor;
 import org.junit.jupiter.api.Test;
 
@@ -37,7 +41,7 @@ public class WorkflowTest {
   }
 
   @Test
-  public void testMockSubWorkflow() {
+  public void testMockTasks() {
     SdkTestingExecutor.Result result =
         SdkTestingExecutor.of(new UberWorkflow())
             .withFixedInput("a", 1)
@@ -48,6 +52,35 @@ public class WorkflowTest {
                 new SumTask(), SumTask.SumInput.create(1L, 2L), SumTask.SumOutput.create(0L))
             .withTaskOutput(
                 new SumTask(), SumTask.SumInput.create(0L, 3L), SumTask.SumOutput.create(0L))
+            .withTaskOutput(
+                new SumTask(), SumTask.SumInput.create(0L, 4L), SumTask.SumOutput.create(42L))
+            .execute();
+
+    assertEquals(42L, result.getIntegerOutput("total"));
+  }
+
+  @Test
+  public void testMockSubWorkflow() {
+    Map<String, Literal> sub1Inputs = new HashMap<>();
+    sub1Inputs.put("left", Literals.ofInteger(1));
+    sub1Inputs.put("right", Literals.ofInteger(2));
+    Map<String, Literal> sub1Outputs = new HashMap<>();
+    sub1Outputs.put("result", Literals.ofInteger(0));
+
+    Map<String, Literal> sub2Inputs = new HashMap<>();
+    sub2Inputs.put("left", Literals.ofInteger(0));
+    sub2Inputs.put("right", Literals.ofInteger(3));
+    Map<String, Literal> sub2Outputs = new HashMap<>();
+    sub2Outputs.put("result", Literals.ofInteger(0));
+
+    SdkTestingExecutor.Result result =
+        SdkTestingExecutor.of(new UberWorkflow())
+            .withFixedInput("a", 1)
+            .withFixedInput("b", 2)
+            .withFixedInput("c", 3)
+            .withFixedInput("d", 4)
+            .withWorkflowOutput(new SubWorkflow(), sub1Inputs, sub1Outputs)
+            .withWorkflowOutput(new SubWorkflow(), sub2Inputs, sub2Outputs)
             .withTaskOutput(
                 new SumTask(), SumTask.SumInput.create(0L, 4L), SumTask.SumOutput.create(42L))
             .execute();
--- a/flytekit-testing/src/main/java/org/flyte/flytekit/testing/DummyTask.java
+++ b/flytekit-testing/src/main/java/org/flyte/flytekit/testing/DummyTask.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2021 Flyte Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.flyte.flytekit.testing;
+
+import java.util.Map;
+import org.flyte.api.v1.Literal;
+import org.flyte.flytekit.SdkRemoteTask;
+import org.flyte.flytekit.SdkType;
+
+public class DummyTask extends SdkRemoteTask<Map<String, Literal>, Map<String, Literal>> {
+
+  private final String name;
+
+  private final SdkType<Map<String, Literal>> inputs;
+
+  private final SdkType<Map<String, Literal>> outputs;
+
+  public DummyTask(
+      String name, SdkType<Map<String, Literal>> inputs, SdkType<Map<String, Literal>> outputs) {
+    this.name = name;
+    this.inputs = inputs;
+    this.outputs = outputs;
+  }
+
+  @Override
+  public String domain() {
+    return "";
+  }
+
+  @Override
+  public String project() {
+    return "";
+  }
+
+  @Override
+  public String name() {
+    return name;
+  }
+
+  @Override
+  public SdkType<Map<String, Literal>> inputs() {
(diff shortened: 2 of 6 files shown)


flytekit-testing/src/main/java/org/flyte/flytekit/testing/DummyTask.java
```diff
@@ -0,0 +1,63 @@
+/*
+ * Copyright 2021 Flyte Authors
```
lupasarin, line 2:
```suggestion
 * Copyright 2022 Flyte Authors
```
Same in the other new files

flytekit-examples/src/test/java/org/flyte/examples/WorkflowTest.java
```diff
@@ -54,4 +58,33 @@ public void testMockSubWorkflow() {
...
+            .withFixedInput("b", 2)
+            .withFixedInput("c", 3)
+            .withFixedInput("d", 4)
+            .withWorkflowOutput(new SubWorkflow(), sub1Inputs, sub1Outputs)
+            .withWorkflowOutput(new SubWorkflow(), sub2Inputs, sub2Outputs)
+            .withTaskOutput(
+                new SumTask(), SumTask.SumInput.create(0L, 4L), SumTask.SumOutput.create(42L))
+            .execute();
+
+    assertEquals(42L, result.getIntegerOutput("total"));
```
lupasarin, line 88: Wondering if there is a way we can test the subworkflows too? i.e. assert on the subworkflow's result

flytekit-testing/src/main/java/org/flyte/flytekit/testing/DummyTask.java
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  @Override
+  public String project() {
+    return "";
+  }
+
+  @Override
+  public String name() {
+    return name;
```
lupasarin, line 51:
```suggestion
    return this.name;
```
Minor, but I would change it for consistency as we use it in inputs and outputs like this.

flytekit-testing/src/main/java/org/flyte/flytekit/testing/SdkTestingExecutor.java
```diff
@@ -289,6 +290,38 @@ public <InputT, OutputT> SdkTestingExecutor withTask(
     return toBuilder().putFixedTask(task.getName(), fixedTask.withRunFn(runFn)).build();
   }
 
+  public <InputT, OutputT> SdkTestingExecutor withWorkflowOutput(
```
lupasarin, line 293: I would suggest to change `withWorkflowOutput` to a better name, since we pass not only the output, but also the workflow, the input and the types.

flytekit-testing/src/main/java/org/flyte/flytekit/testing/SdkTestingExecutor.java
```diff
@@ -289,6 +290,38 @@ public <InputT, OutputT> SdkTestingExecutor withTask(
...
+      InputT input,
+      SdkType<OutputT> outputType,
+      OutputT output) {
+    return withWorkflowOutput(
+        workflow, inputType.toLiteralMap(input), outputType.toLiteralMap(output));
+  }
+
+  public SdkTestingExecutor withWorkflowOutput(
+      SdkWorkflow workflow, Map<String, Literal> input, Map<String, Literal> output) {
+    // TODO type checking
```
lupasarin, line 305: Is this something to be done in this PR or in a future one?

flytekit-testing/src/main/java/org/flyte/flytekit/testing/SdkTestingExecutor.java
```diff
@@ -289,6 +290,38 @@ public <InputT, OutputT> SdkTestingExecutor withTask(
...
+    // fixed tasks
+    SdkType<Map<String, Literal>> inputType = TestingSdkType.of(intf.inputs());
+    SdkType<Map<String, Literal>> outputType = TestingSdkType.of(intf.outputs());
+    TestingRunnableTask<Map<String, Literal>, Map<String, Literal>> fixedTask =
+        getFixedTaskOrDefault(workflow.getName(), inputType, outputType);
+
+    // replace workflow
+    SdkWorkflowBuilder builder = new SdkWorkflowBuilder();
+    SdkWorkflow w = new TestingWorkflow(workflow, inputType, outputType);
+    w.expand(builder);
```
lupasarin, line 317: I find it a bit confusing that `w` is just used here to expand and then never used afterwards. Only the builder is used later on, which is created empty and personally to be it is not clear what kind of "magic" expand will do.

flytekit-testing/src/main/java/org/flyte/flytekit/testing/SdkTestingExecutor.java
```diff
@@ -289,6 +290,38 @@ public <InputT, OutputT> SdkTestingExecutor withTask(
...
+
+    // fixed tasks
+    SdkType<Map<String, Literal>> inputType = TestingSdkType.of(intf.inputs());
+    SdkType<Map<String, Literal>> outputType = TestingSdkType.of(intf.outputs());
+    TestingRunnableTask<Map<String, Literal>, Map<String, Literal>> fixedTask =
+        getFixedTaskOrDefault(workflow.getName(), inputType, outputType);
+
+    // replace workflow
+    SdkWorkflowBuilder builder = new SdkWorkflowBuilder();
+    SdkWorkflow w = new TestingWorkflow(workflow, inputType, outputType);
```
lupasarin, line 316:
```suggestion
    SdkWorkflow testingWorkflow = new TestingWorkflow(workflow, inputType, outputType);
```
I think it is clearer to differentiate it from `workflow` variable.

flytekit-testing/src/main/java/org/flyte/flytekit/testing/TestingWorkflow.java
```diff
@@ -0,0 +1,58 @@
...
+    this.inputs = inputs;
+    this.outputs = outputs;
+  }
+
+  @Override
+  public void expand(SdkWorkflowBuilder builder) {
+    @Var SdkTransform task = new DummyTask(workflow.getName(), inputs, outputs);
+    for (Map.Entry<String, Variable> i : inputs.getVariableMap().entrySet()) {
+      task =
+          task.withInput(i.getKey(), builder.inputOf(i.getKey(), i.getValue().literalType(), ""));
```
lupasarin, line 49: I want to try and see if we can avoid passing i.getKey() twice.

lupasarin: One more note on this: it looks like the same `builder` is used in every iteration and it is not cleaned up. Therefore, I would expect that the inputs get accumulated and each iteration also assigns the previous values, e.g.
```
task = task.withInput(i.getKey(), builder.inputOf(i.getKey(), i.getValue().literalType(), ""));
```
with `i = 0` saves with key0 = builder(key0, value0) and the builder saves key0,value0
with `i=1` saves with key1 = builder(key1, value1) and the builder saves key1,value1, but the builder already contained key0,value0 so we also save key0,value0. Does that make sense?
I would like to confirm if this is expected behavior.

flytekit-examples/src/test/java/org/flyte/examples/WorkflowTest.java
narape: For the time being, we can only assert on result values. However in this test we are mocking the subworkflow, meaning that instead of running the subworkflow we use the supplied output

flytekit-testing/src/main/java/org/flyte/flytekit/testing/SdkTestingExecutor.java
narape: I think that the name is a good fit.

It is following the lead of other methods in this class, like `withTaskOutput`. 

The whole idea of this method is that the workflow will run, but:
*  it will skip the workflow (sub workflow as a workflow inside a workflow) execution
* use the supplied output instead
* but only if the input to the workflow is equals to the supplied one

This is equivalent to an hypothetical workflow api and mocking with Mockito:
```
Mockito.when(workflow.runSubWorkflow(workflow, input)
            .thenReturn(output)

File: flytekit-testing/src/main/java/org/flyte/flytekit/testing/SdkTestingExecutor.java [hunk 9]
```diff
@@ -289,6 +290,38 @@ public <InputT, OutputT> SdkTestingExecutor withTask(
...
+      SdkWorkflow workflow,
+      SdkType<InputT> inputType,
+      InputT input,
+      SdkType<OutputT> outputType,
+      OutputT output) {
+    return withWorkflowOutput(
+        workflow, inputType.toLiteralMap(input), outputType.toLiteralMap(output));
+  }
+
+  public SdkTestingExecutor withWorkflowOutput(
```
narape, line 303: So far, users interact with flytekit-java type system by using `SdkType` so I would rather have only the method exposed in line 293

flytekit-testing/src/main/java/org/flyte/flytekit/testing/TestingWorkflow.java
narape: The caller of `expand` method in is charge of supplying a brand new builder every time.

flytekit-testing/src/main/java/org/flyte/flytekit/testing/SdkTestingExecutor.java
narape, line 293: I would like to have an unit test for this new functionality

narape (commented):
Thanks for your contribution @RustedBones 

My main concern is exposing `Literal` in the user interface. I would create a new branch based on yours to work on this

narape:
@RustedBones I created a new PR(https://github.com/flyteorg/flytekit-java/pull/123) based on yours. There I managed to provide (sub)workflow mocking but using SdkTypes.

narape:
https://github.com/flyteorg/flytekit-java/pull/123 was merged and supersedes this one by relying only on SdkTypes

narape:
@RustedBones thanks by the way for starting this

PR #124, opened 2022-08-02
Don't include module artifacts in jflyte

narape:
# TL;DR
Revert the dependency of jflyte and it's modules

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [ ] Unit tests added

## Complete description
On we made a mistake to create a dependency from jflyte to its modules, meaning that we include the modules artifacts and its dependencies inside jflye, defeating the purpose of independent modules.

Notice that the inverse is correct, a module should depend on jflyte, as it would implement the interface `FileSystem` or `ArtifactStager` defined in jflyte.

Beside possible dependency conflicts of including modules artifact in jflyte, it also has the unexpected consequence of duplicating the artifacts in the docker image, as modules have their own independent modules.

### DIFF
--- a/jflyte/pom.xml
+++ b/jflyte/pom.xml
@@ -58,10 +58,16 @@
     <dependency>
       <groupId>org.flyte</groupId>
       <artifactId>jflyte-aws</artifactId>
+      <!-- We want to have the dependency, so it is already built before building the docker image,
+           but we want to avoid the artifacts from the module to pollute jflyte -->
+      <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>org.flyte</groupId>
       <artifactId>jflyte-google-cloud</artifactId>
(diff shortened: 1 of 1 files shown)


lupasarin (approved):

jflyte/pom.xml
```diff
@@ -58,10 +58,16 @@
     <dependency>
       <groupId>org.flyte</groupId>
       <artifactId>jflyte-aws</artifactId>
+      <!-- We want to have the dependency, so it is already built before building the docker image,
+           but we want to avoid the artifacts from the module to pollute jflyte -->
+      <scope>provided</scope>
```
ckiosidis, line 63: 👍

PR #129, opened 2022-08-22
Mock launchplans

narape:
# TL;DR
Be able to stub/fake launchplans during unit testing

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [] Unit tests added

## Complete description
Add the capability of replacing the execution of launchplans for a value or function when executing an unit test.

## Follow-up issue
_NA_

### DIFF
--- a/flytekit-local-engine/src/main/java/org/flyte/localengine/ExecutionContext.java
+++ b/flytekit-local-engine/src/main/java/org/flyte/localengine/ExecutionContext.java
@@ -30,7 +30,7 @@ public abstract class ExecutionContext {
 
   abstract Map<String, DynamicWorkflowTask> dynamicWorkflowTasks();
 
-  abstract Map<String, WorkflowTemplate> workflows();
+  abstract Map<String, WorkflowTemplate> workflowTemplates();
 
   abstract Map<String, RunnableLaunchPlan> runnableLaunchPlans();
 
@@ -42,7 +42,7 @@ public abstract class ExecutionContext {
         .executionListener(NoopExecutionListener.create())
         .runnableTasks(emptyMap())
         .runnableLaunchPlans(emptyMap())
-        .workflows(emptyMap());
+        .workflowTemplates(emptyMap());
   }
 
   @AutoValue.Builder
@@ -53,7 +53,7 @@ public abstract class ExecutionContext {
     public abstract Builder dynamicWorkflowTasks(
         Map<String, DynamicWorkflowTask> dynamicWorkflowTasks);
 
-    public abstract Builder workflows(Map<String, WorkflowTemplate> workflows);
+    public abstract Builder workflowTemplates(Map<String, WorkflowTemplate> workflowTemplates);
 
     public abstract Builder runnableLaunchPlans(
         Map<String, RunnableLaunchPlan> runnableLaunchPlans);
--- a/flytekit-local-engine/src/main/java/org/flyte/localengine/ExecutionContext.java
+++ b/flytekit-local-engine/src/main/java/org/flyte/localengine/ExecutionContext.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2021 Flyte Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.flyte.localengine;
+
+import static java.util.Collections.emptyMap;
+
+import com.google.auto.value.AutoValue;
+import java.util.Map;
+import org.flyte.api.v1.DynamicWorkflowTask;
+import org.flyte.api.v1.RunnableTask;
+import org.flyte.api.v1.WorkflowTemplate;
+
+@AutoValue
+public abstract class ExecutionContext {
+  abstract Map<String, RunnableTask> runnableTasks();
+
+  abstract Map<String, DynamicWorkflowTask> dynamicWorkflowTasks();
+
+  abstract Map<String, WorkflowTemplate> workflows();
+
+  abstract Map<String, RunnableLaunchPlan> runnableLaunchPlans();
+
+  abstract ExecutionListener executionListener();
+
+  public static Builder builder() {
+    return new AutoValue_ExecutionContext.Builder()
+        .dynamicWorkflowTasks(emptyMap())
+        .executionListener(NoopExecutionListener.create())
+        .runnableTasks(emptyMap())
+        .runnableLaunchPlans(emptyMap())
+        .workflows(emptyMap());
+  }
+
+  @AutoValue.Builder
+  public abstract static class Builder {
+
+    public abstract Builder runnableTasks(Map<String, RunnableTask> runnableTasks);
+
+    public abstract Builder dynamicWorkflowTasks(
+        Map<String, DynamicWorkflowTask> dynamicWorkflowTasks);
+
+    public abstract Builder workflows(Map<String, WorkflowTemplate> workflows);
+
+    public abstract Builder runnableLaunchPlans(
+        Map<String, RunnableLaunchPlan> runnableLaunchPlans);
+
+    public abstract Builder executionListener(ExecutionListener executionListener);
+
+    public abstract ExecutionContext build();
+  }
+}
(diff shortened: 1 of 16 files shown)


flytekit-local-engine/src/main/java/org/flyte/localengine/ExecutionNodeCompiler.java
```diff
@@ -54,35 +57,26 @@
  * </ul>
  */
 class ExecutionNodeCompiler {
+  private final ExecutionContext executionContext;
+
+  ExecutionNodeCompiler(ExecutionContext executionContext) {
+    this.executionContext = executionContext;
```
narape, line 63: add requiresNonNull so we catch bugs closer to the source

flytekit-local-engine/src/main/java/org/flyte/localengine/LocalEngine.java
```diff
@@ -37,115 +36,42 @@
...
-  public static Map<String, Literal> compileAndExecute(
-      WorkflowTemplate template,
-      Map<String, RunnableTask> runnableTasks,
-      Map<String, DynamicWorkflowTask> dynamicWorkflowTasks,
-      Map<String, Literal> inputs,
-      ExecutionListener listener) {
-    return compileAndExecute(
-        template, runnableTasks, dynamicWorkflowTasks, emptyMap(), inputs, listener);
+  public LocalEngine(ExecutionContext context) {
+    this.context = context;
```
narape, line 42: add requiresNonNull so we catch bugs closer to the source

flytekit-testing/src/main/java/org/flyte/flytekit/testing/TestingRunnableNode.java
```diff
@@ -0,0 +1,107 @@
...
+
+    if (fixedOutputs.containsKey(input)) {
+      return outputType.toLiteralMap(fixedOutputs.get(input));
+    }
+
+    if (runFn == null) {
+      String message =
+          String.format(
+              "Can't find input %s for remote task [%s] across known task inputs, "
+                  + "use SdkTestingExecutor#withTaskOutput or SdkTestingExecutor#withTask",
```
narape, line 83: This string is not generic for both sub clases

flytekit-local-engine/src/main/java/org/flyte/localengine/ExecutionContext.java
```diff
@@ -0,0 +1,65 @@
...
+import org.flyte.api.v1.RunnableTask;
+import org.flyte.api.v1.WorkflowTemplate;
+
+@AutoValue
+public abstract class ExecutionContext {
+  abstract Map<String, RunnableTask> runnableTasks();
+
+  abstract Map<String, DynamicWorkflowTask> dynamicWorkflowTasks();
+
+  abstract Map<String, WorkflowTemplate> workflows();
```
lupasarin, line 33: Maybe rename it to `workflowTemplates`?

flytekit-local-engine/src/main/java/org/flyte/localengine/ExecutionNodeCompiler.java
```diff
@@ -99,53 +93,83 @@ static ExecutionNode compile(
...
-        default:
-          throw new IllegalArgumentException(
-              String.format("Unsupported Reference.Kind: [%s]", reference.kind()));
-      }
-    } else {
-      // this must be a task
-      String taskName = node.taskNode().referenceId().name();
+      return compileWorkflowNode(node, upstreamNodeIds);
+    } else if (node.taskNode() != null) {
+      return compileTaskNode(node, upstreamNodeIds);
```
lupasarin, line 98: I think we should pass `node.taskNode()` as first param, instead of node, since no other methods in node are called in compileTaskNode.

flytekit-local-engine/src/main/java/org/flyte/localengine/ExecutionNodeCompiler.java
```diff
@@ -99,53 +93,83 @@ static ExecutionNode compile(
...
+        String.format("Node [%s] must be a task, branch or workflow node", node.id()));
+  }
 
-      DynamicWorkflowTask dynamicWorkflowTask = dynamicWorkflowTasks.get(taskName);
-      if (dynamicWorkflowTask != null) {
+  private ExecutionNode compileWorkflowNode(Node node, List<String> upstreamNodeIds) {
+    WorkflowNode.Reference reference = node.workflowNode().reference();
+    switch (reference.kind()) {
+      case SUB_WORKFLOW_REF:
+        return compileSubWorkflowRef(node, upstreamNodeIds, reference);
```
lupasarin, line 109: We can pass `reference.subWorkflowRef().name()` as third parameter and call it subworkflowName. Since it is the only thing that compileSubWorkflowRef needs from reference.

flytekit-local-engine/src/main/java/org/flyte/localengine/ExecutionNodeCompiler.java
```diff
@@ -99,53 +93,83 @@ static ExecutionNode compile(
...
+      case LAUNCH_PLAN_REF:
+        return compileLaunchPlanRef(node, upstreamNodeIds, reference);
```
lupasarin, line 111: We can pass `reference.launchPlanRef().name()` as third parameter and call it launchPlanName. Since it is the only thing that compileLaunchPlanRef needs from reference.

flytekit-local-engine/src/test/java/org/flyte/localengine/ExecutionNodeCompilerTest.java
```diff
@@ -205,7 +204,7 @@ private static ExecutionNode createExecutionNode(String nodeId, List<String> ups
     return ExecutionNode.builder()
         .nodeId(nodeId)
         .upstreamNodeIds(upstreamNodeIds)
-        .runnableTask(new EmptyRunnableTask())
+        .runnableNode(new EmptyRunnableTask())
```
lupasarin, line 207: Is it ok to pass EmptyRunnableTask here since it is a generic createExecutionNode method?

flytekit-testing/src/main/java/org/flyte/flytekit/testing/TestingRunnableNode.java
```diff
@@ -0,0 +1,107 @@
...
+        IdT extends PartialIdentifier,
+        InputT,
+        OutputT,
+        T extends TestingRunnableNode<IdT, InputT, OutputT, T>>
+    implements RunnableNode {
+  protected final IdT id;
+  protected final SdkType<InputT> inputType;
+  protected final SdkType<OutputT> outputType;
+
+  // @Nullable
```
lupasarin, line 37:
```suggestion
```

flytekit-local-engine/src/main/java/org/flyte/localengine/RunnableLaunchPlan.java
```diff
@@ -0,0 +1,21 @@
+/*
```
lupasarin, line 1: A test for this new class?

lupasarin (approved):

flytekit-local-engine/src/main/java/org/flyte/localengine/ExecutionNodeCompiler.java
narape:
```java
    return ExecutionNode.builder()
        .nodeId(node.id())
        .bindings(node.inputs())
        .runnableNode(runnableTask)
        .upstreamNodeIds(upstreamNodeIds)
        .attempts(attempts)
        .build();
```
Actually we use it when creating the execution node

lupasarin: True, I thought that wasn't part of the method anymore, but a separate one

flytekit-testing/src/main/java/org/flyte/flytekit/testing/TestingRunnableNode.java
narape: I added a explanation of why this line is commented oput

flytekit-testing/src/main/java/org/flyte/flytekit/testing/TestingRunnableNode.java
narape: We need it but the code is not obvious so it is not easy to understand why. I will refactor it a bit

flytekit-local-engine/src/main/java/org/flyte/localengine/RunnableLaunchPlan.java
narape: This is an interface with no methods, what would you test here?

flytekit-testing/src/main/java/org/flyte/flytekit/testing/TestingRunnableLaunchPlan.java
```diff
@@ -0,0 +1,48 @@
+/*
```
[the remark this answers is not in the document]
narape: I added a test for `TestingRunnableNode` instead, that is the super class that has all the interesting logic. This class is very minimal, do you think that we must test it?

PR #132, opened 2022-08-26
Support branch nodes in the local execution engine

narape:
# TL;DR
Support branch nodes in the local execution engine (and therefore in unit testing)

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [ ] Smoke tested
 - [X] Unit tests added

## Complete description
Before this PR, workflows that contains branch nodes couldn't be unit tested, they would throw `IllegalArgumentException` with the message `BranchNode isn't yet supported for local execution`. Now the local engine is able to evaluate the boolean conditions in the `IfBlock` and execute the corresponding node.

## Tracking Issue
_NA_

## Follow-up issue
_NA_


flytekit-local-engine/src/main/java/org/flyte/localengine/ExecutionNodeCompiler.java
```diff
@@ -106,7 +107,41 @@ private static List<String> compileUpstreamNodeIds(Node node) {
...
+    IfElseBlock ifElseBlock = node.branchNode().ifElse();
+
+    List<ExecutionIfBlock> ifBlocks =
+        Stream.concat(Stream.of(ifElseBlock.case_()), ifElseBlock.other().stream())
+            .map(
+                ifBlock ->
+                    ExecutionIfBlock.create(ifBlock.condition(), compile(ifBlock.thenNode())))
+            .collect(toList());
+    ExecutionNode elseNode = compileIfNotNull(ifElseBlock.elseNode());
+    // XXX support node error
```
narape, line 121: remove

flytekit-local-engine/src/main/java/org/flyte/localengine/LocalEngine.java
```diff
@@ -50,10 +51,20 @@ public Map<String, Literal> compileAndExecute(
     return execute(executionNodes, inputs, template.outputs());
   }
 
+  public boolean isRecoverable(Throwable e) {
```
narape, line 54: Move back

flytekit-local-engine/src/test/java/org/flyte/localengine/examples/TestCaseExhaustivenessWorkflow.java
```diff
@@ -0,0 +1,84 @@
...
+import com.google.auto.service.AutoService;
+import com.google.auto.value.AutoValue;
+import org.flyte.flytekit.SdkBindingData;
+import org.flyte.flytekit.SdkRunnableTask;
+import org.flyte.flytekit.SdkTransform;
+import org.flyte.flytekit.SdkWorkflow;
+import org.flyte.flytekit.SdkWorkflowBuilder;
+import org.flyte.flytekit.jackson.JacksonSdkType;
+
+// 3X+1
```
narape, line 32: change comment

flytekit-local-engine/src/main/java/org/flyte/localengine/ExecutionNodeCompiler.java
```diff
@@ -106,7 +107,40 @@ private static List<String> compileUpstreamNodeIds(Node node) {
...
+    assert node.branchNode() != null;
+    IfElseBlock ifElseBlock = node.branchNode().ifElse();
+
+    List<ExecutionIfBlock> ifBlocks =
+        Stream.concat(Stream.of(ifElseBlock.case_()), ifElseBlock.other().stream())
+            .map(
+                ifBlock ->
+                    ExecutionIfBlock.create(ifBlock.condition(), compile(ifBlock.thenNode())))
+            .collect(toList());
+    ExecutionNode elseNode = compileIfNotNull(ifElseBlock.elseNode());
```
ckiosidis, line 120: what is the use case for building a null elseNode?

flytekit-local-engine/src/main/java/org/flyte/localengine/ExecutionBranchNode.java
```diff
@@ -0,0 +1,62 @@
...
+  @AutoValue.Builder
+  public abstract static class Builder {
+
+    public abstract Builder ifNodes(List<ExecutionIfBlock> ifNodes);
+
+    public abstract Builder elseNode(ExecutionNode elseNode);
+
+    public abstract Builder error(NodeError error);
+
+    abstract ExecutionBranchNode autoBuild();
```
ckiosidis, line 52: how does this autoBuild work?

flytekit-local-engine/src/test/java/org/flyte/localengine/examples/CollatzConjectureStepWorkflow.java
```diff
@@ -0,0 +1,52 @@
...
+    SdkBindingData isOdd =
+        builder.apply("is_odd", new IsEvenTask().withInput("x", x)).getOutput("res");
+
+    SdkBindingData nextX =
+        builder
+            .apply(
+                "decide",
+                when(
+                        "was_even",
+                        eq(ofBoolean(true), isOdd),
```
ckiosidis, line 45: flytekit python provides some helper is_true() methods to evaluate boolean conditions easier (when the value is boolean as well) 
Maybe something that we could do as well to simplify this syntax? 
https://github.com/flyteorg/flytekit/blob/952eb9fb5f5dfd3c1c31d9e3826d70d73f5c1568/tests/flytekit/unit/core/test_workflows.py#L77

flytekit-local-engine/src/main/java/org/flyte/localengine/ExecutionNodeCompiler.java
narape: There is a error node, meaning that it would throw an exception if all if clauses in the change were false

flytekit-local-engine/src/main/java/org/flyte/localengine/BooleanExpressionEvaluator.java
```diff
@@ -0,0 +1,233 @@
...
+        return integerEq(left, right);
+      case FLOAT_VALUE:
+        return floatEq(left, right);
+      default:
+        return Objects.equals(left, right);
+    }
+  }
+
+  private static boolean integerEq(Primitive left, Primitive right) {
+    assert left.kind() == Primitive.Kind.INTEGER_VALUE;
```
ckiosidis, line 104: we used Verify previously https://github.com/flyteorg/flytekit-java/blob/3cf3ee4ac27a3e1bad231be2fa9a30b8afa679ed/jflyte/src/main/java/org/flyte/jflyte/FlyteAdminClient.java#L175
should we keep them consistent?

flytekit-local-engine/src/test/java/org/flyte/localengine/examples/TestCaseExhaustivenessWorkflow.java
```diff
@@ -0,0 +1,83 @@
...
+                "decide",
+                when("eq_1", eq(ofInteger(1L), x), NoOp.of(x))
+                    .when("eq_2", eq(ofInteger(2L), x), NoOp.of(x)))
+            .getOutput("x");
+
+    builder.output("nextX", nextX);
+  }
+
+  @AutoService(SdkRunnableTask.class)
+  public static class NoOp extends SdkRunnableTask<NoOpInput, NoOpOutput> {
```
ckiosidis, line 50: add a comment on what this class task is

flytekit-local-engine/src/test/java/org/flyte/localengine/examples/CollatzConjectureStepWorkflow.java
```diff
@@ -0,0 +1,52 @@
...
+import static org.flyte.flytekit.SdkConditions.when;
+
+import com.google.auto.service.AutoService;
+import org.flyte.flytekit.SdkBindingData;
+import org.flyte.flytekit.SdkWorkflow;
+import org.flyte.flytekit.SdkWorkflowBuilder;
+
+// if x is even, then x/2 else 3x+1
+@AutoService(SdkWorkflow.class)
+public class CollatzConjectureStepWorkflow extends SdkWorkflow {
```
ckiosidis, line 31: should we move the tasks inside this file for simplicity of the example?

flytekit-local-engine/src/main/java/org/flyte/localengine/ExecutionBranchNode.java
narape: This allows me to create the build method to validate that either elseNode or error node should be set. 
See https://github.com/google/auto/blob/master/value/userguide/builders-howto.md#-validate-property-values

flytekit-local-engine/src/main/java/org/flyte/localengine/ExecutionBranchNode.java
```diff
@@ -0,0 +1,62 @@
...
+
+    public final ExecutionBranchNode build() {
+      ExecutionBranchNode node = autoBuild();
+      if (node.elseNode() == null && node.error() == null) {
```
narape, line 56: Actually I missed the case that both are not null.

flytekit-local-engine/src/test/java/org/flyte/localengine/examples/CollatzConjectureStepWorkflow.java
narape: Done

flytekit-local-engine/src/test/java/org/flyte/localengine/examples/TestCaseExhaustivenessWorkflow.java
narape: This is a class just for testing that I can reach all the combinations. I moved this workflow as inner class of the test to avoid documenting it. 
Is this OK with you?

flytekit-local-engine/src/main/java/org/flyte/localengine/BooleanExpressionEvaluator.java
narape: The local engine doesn't have a dependency with guava, but this assertion should always pass as it is a private method and we will always call it aster verifying that the primitive is an integer

```java
    switch (left.kind()) {
      case INTEGER_VALUE:
        return integerEq(left, right);
      case FLOAT_VALUE:
        return floatEq(left, right);
      default:
        return Objects.equals(left, right);
    }
```

I would rather remove the assertion than adding guava as dependency to this module.
What do you say @ckiosidis

flytekit-local-engine/src/test/java/org/flyte/localengine/examples/CollatzConjectureStepWorkflow.java
narape: I added `isTrue` to a `SdkConditions` so it is used as `isTrue(isOdd)` instead of what I think that you are telling me `isOdd.isTrue()`. 

To support the second case I think we will need to decide carefully what we want to put there so it is better for a new PR

flytekit-local-engine/src/main/java/org/flyte/localengine/BooleanExpressionEvaluator.java
ckiosidis: ok let's not add guava

flytekit-local-engine/src/main/java/org/flyte/localengine/ExecutionBranchNode.java
```diff
@@ -53,8 +53,15 @@ public abstract static class Builder {
...
       ExecutionBranchNode node = autoBuild();
-      if (node.elseNode() == null && node.error() == null) {
-        throw new IllegalStateException("Both elseNode and errorNode cannot be nulls");
+
+      if (node.ifNodes().isEmpty()) {
+        throw new IllegalStateException(
+            "There must be at least one if-then node");
+      }
+
+      if ((node.elseNode() == null) == (node.error() == null)) {
```
ckiosidis, line 61: usage of == here can be misleading

flytekit-local-engine/src/test/java/org/flyte/localengine/examples/CollatzConjectureStepWorkflow.java
```diff
@@ -45,7 +44,7 @@ public void expand(SdkWorkflowBuilder builder) {
                 "decide",
                 when(
                         "was_even",
-                        eq(ofBoolean(true), isOdd),
+                        isTrue(isOdd),
```
ckiosidis, line 47: that's good enough i believe, no need to complicate it more 👍

flytekit-examples/src/main/java/org/flyte/examples/ConditionalGreetingWorkflow.java
```diff
@@ -0,0 +1,45 @@
...
+
+  @Override
+  public void expand(SdkWorkflowBuilder builder) {
+    SdkBindingData name = builder.inputOfString("name");
+    SdkBindingData greeting =
+        builder
+            .apply(
+                "decide",
+                SdkConditions.when(
+                        "when-empty", eq(name, ofString("")), GreetTask.of(ofString("World")))
```
ckiosidis, line 39: eq and other operators do not support primitives?

flytekit-examples/src/test/java/org/flyte/examples/WorkflowTest.java
```diff
@@ -85,4 +87,15 @@ public void testMockSubWorkflow() {
 
     assertEquals(15L, result.getIntegerOutput("total"));
   }
+
+  @ParameterizedTest
+  @CsvSource({"'','Welcome, World!'", "Joe,'Welcome, Joe!'"})
```
ckiosidis, line 92: using a parameterized test in this example adds a bit of complexity for users i think. It could be a straight forward simple test instead, so it's easier to copy/paste. just an observation

ckiosidis (approved):

PR #134, opened 2022-09-22
reuse test container locally

ckiosidis:
Signed-off-by: Babis Kiosidis <ckiosidis@gmail.com>
# TL;DR

Set the config in your local environment to reuse the flyte sandbox test container throughout the tests. The config can only be set in the home directory

```
 cat ~/.testcontainers.properties 
testcontainers.reuse.enable=true
```

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
 _How did you fix the bug, make the feature etc. Link to any design docs etc_

## Tracking Issue

## Follow-up issue
_NA_
OR
_https://github.com/flyteorg/flyte/issues/<number>_

### DIFF
--- a/integration-tests/src/test/java/org/flyte/SerializeJavaIT.java
+++ b/integration-tests/src/test/java/org/flyte/SerializeJavaIT.java
@@ -21,21 +21,25 @@ import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.equalTo;
 
 import java.io.File;
+import java.nio.file.Path;
 import java.util.stream.Stream;
-import org.flyte.utils.FlyteSandboxClient;
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.TempDir;
 
 public class SerializeJavaIT {
   private static final String CLASSPATH = "flytekit-examples/target/lib";
-  @TempDir File tempDir;
+
+  @TempDir Path managed;
 
   @Test
-  @Disabled
   public void testSerializeWorkflows() {
     try {
-      CLIENT.serializeWorkflows(CLASSPATH, tempDir.getAbsolutePath());
+      File current = new File("target/protos");
+      current.mkdir();
+
+      File tempDir = managed.resolve(current.getAbsolutePath()).toFile();
+
+      CLIENT.serializeWorkflows(CLASSPATH, tempDir.getPath());
 
       boolean hasFibonacciWorkflow =
           Stream.of(tempDir.list())
--- a/integration-tests/src/test/java/org/flyte/SerializeJavaIT.java
+++ b/integration-tests/src/test/java/org/flyte/SerializeJavaIT.java
@@ -16,6 +16,7 @@
  */
 package org.flyte;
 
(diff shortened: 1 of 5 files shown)


integration-tests/src/test/java/org/flyte/utils/FlyteSandboxContainer.java
```diff
@@ -35,7 +35,7 @@ public class FlyteSandboxContainer extends GenericContainer<FlyteSandboxContaine
 
   public static final String IMAGE_NAME = "ghcr.io/flyteorg/flyte-sandbox:v1.1.0";
 
-  public static final FlyteSandboxContainer INSTANCE = new FlyteSandboxContainer();
+  public static final FlyteSandboxContainer INSTANCE = new FlyteSandboxContainer().withReuse(true);
```
narape, line 38:
```suggestion
  public static final FlyteSandboxContainer INSTANCE = new FlyteSandboxContainer()
          // Note to the developer: to enable test container reuse, please do the following
          // echo testcontainers.reuse.enable=true > ~/.testcontainers.properties
         .withReuse(true);
```

narape (approved):
I added a suggestion to add a comment as it would be easier to find that this PR description

integration-tests/src/test/java/org/flyte/utils/FlyteSandboxContainer.java
narape: Maybe reformatting is needed after applying my sugestion

integration-tests/src/test/java/org/flyte/SerializeJavaIT.java
```diff
@@ -21,21 +21,25 @@
...
 
   @Test
-  @Disabled
   public void testSerializeWorkflows() {
     try {
-      CLIENT.serializeWorkflows(CLASSPATH, tempDir.getAbsolutePath());
+      File current = new File("target/protos");
+      current.mkdir();
+
+      File tempDir = managed.resolve(current.getAbsolutePath()).toFile();
```
narape, line 40: I don't understand why you make `mkdir` first and then resolve. In my head this makes more sense

```java
  @TempDir File tempDir;

//...

    try {
      File protosDir = managed.resolve("target/protos").toFile();
      protosDir.mkdir();
      CLIENT.serializeWorkflows(CLASSPATH, protosDir.getPath());
...

Comment (ckiosidis) [reply integration-tests/src/test/java/org/flyte/SerializeJavaIT.java]: this wont work because the default location for a @TempDir is somewhere in a tmp /var/... and the docker images that we run do not have access to that path.
We need to first create the File that points in the project's path, then we pass it to the tempdir

```
[docker-java-stream--796442191] INFO 🐳 [ghcr.io/flyteorg/flytekit-java:0.3.20-SNAPSHOT] - STDERR: Caused by: java.io.FileNotFoundException: /var/folders/vl/46bdmk_x37s3vd61fcfp05f80000gn/T/junit15098090948395355541/target/protos/00_org.flyte.examples.PrintMessageTask_1.pb (No such file or directory)
```

PR #145, opened 2022-11-04
Child only resource discovery

honnix:
# TL;DR
Configure child only resource discovery to avoid discovering certain resources across class loaders.

## Type
 - [x] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
For certain resources such as `org/slf4j/impl/StaticLoggerBinder.class`, it does not make sense to discovery a parent one because it will almost certain cause classloading problem when mixing parent class and child class.

## Tracking Issue
Closes https://github.com/flyteorg/flyte/issues/3051

## Follow-up issue
_NA_

### DIFF
--- a/jflyte/src/main/java/org/flyte/jflyte/ChildFirstClassLoader.java
+++ b/jflyte/src/main/java/org/flyte/jflyte/ChildFirstClassLoader.java
@@ -24,6 +24,9 @@ import java.util.ArrayList;
 import java.util.Enumeration;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 /**
  * {@link URLClassLoader} that loads classes into child class loader, instead of parent.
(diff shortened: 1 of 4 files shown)


jflyte/src/main/java/org/flyte/jflyte/ChildFirstClassLoader.java
```diff
@@ -42,6 +45,9 @@ class ChildFirstClassLoader extends URLClassLoader {
   private static final String[] PARENT_FIRST_PACKAGE_PREFIXES =
       new String[] {"org.flyte.api.v1.", "org.flyte.jflyte.api."};
 
+  private static final Set<String> CHILD_ONLY_RESOURCES =
```
honnix, line 48: This project is still targeting Java 8, so no factory method from `Set` to use.

rafaelraposospot (approved):

honnix:
I don't have permission to merge. Maybe I should be added to some team?

PR #152, opened 2022-11-26
Block more resource loading to parent

honnix:
# TL;DR
Block service resource loading to parent.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
`ServiceLoader` tries to find resources to understand what classes to load, and for the following scenario, it will fail with a mysteries error message.

* `com.example.Foo` exists both in parent and child classpath, in two different jars
* `com.example.FooImpl` only exists in parent classpath
* Task code (under child classpath) uses `ServiceLoader` to discover implementation of `com.example.Foo`, using `ChildFirstClassLoader`
* `ServiceLoader` uses `ChildFirstClassLoader` to find resources from both child and parent classpath, and in this case `com.example.FooImpl` is discovered
* `ServiceLoader` verifies whether `com.example.FooImpl` (loaded by parent class loader, and its super class `com.example.Foo` is also loaded by parent class loader; remember there are two copies of `com.example.Foo`?) can be assigned to `com.example.Foo` (loaded by child class loader)
* The verification fails because there are two copies of `com.example.Foo` loaded by different class loaders, with an error message like "java.util.ServiceConfigurationError: com.example.Foo: com.example.FooImpl not a subtype"

## Tracking Issue
Closes https://github.com/flyteorg/flyte/issues/3103

## Follow-up issue
_NA_

### DIFF
--- a/integration-tests/src/test/java/org/flyte/JavaExamplesIT.java
+++ b/integration-tests/src/test/java/org/flyte/JavaExamplesIT.java
@@ -23,6 +23,7 @@ import static org.hamcrest.Matchers.equalTo;
 
 import flyteidl.core.Literals;
 import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.TestInstance;
 import org.testcontainers.shaded.com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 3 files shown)


integration-tests/src/test/java/org/flyte/JavaExamplesIT.java
```diff
@@ -63,6 +64,7 @@ public void testFibonacciWorkflow() {
   }
 
   @Test
+  @Disabled
```
honnix, line 67: For some reason this always times out.

narape (approved):

PR #157, opened 2022-12-12
Staging in parallel

honnix:
# TL;DR
Staging artifacts in parallel.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
Currently artifacts are staged sequentially, which is very slow due to IO blocking. By staging them in multiple threads, the throughput can be largely improved.

## Follow-up issue
_NA_

### DIFF
--- a/jflyte/src/main/java/org/flyte/jflyte/ArtifactStager.java
+++ b/jflyte/src/main/java/org/flyte/jflyte/ArtifactStager.java
@@ -37,6 +37,10 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.stream.Collectors;
 import org.flyte.jflyte.api.FileSystem;
 import org.flyte.jflyte.api.Manifest;
 import org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


jflyte/src/main/java/org/flyte/jflyte/RegisterWorkflows.java
```diff
@@ -73,7 +74,8 @@ public Integer call() {
 
     try (FlyteAdminClient adminClient =
         FlyteAdminClient.create(config.platformUrl(), config.platformInsecure(), tokenSource)) {
-      Supplier<ArtifactStager> stagerSupplier = () -> ArtifactStager.create(config, modules);
+      Supplier<ArtifactStager> stagerSupplier =
+          () -> ArtifactStager.create(config, modules, new ForkJoinPool());
```
honnix, line 78: It seems good enough to use default configuration of `ForkJoinPool`.

jflyte/src/main/java/org/flyte/jflyte/ArtifactStager.java
```diff
@@ -116,6 +147,7 @@ void stageArtifact(Artifact artifact, ByteSource content) {
         throw new UncheckedIOException(e);
       }
     } else {
+      LOG.info("[{}] already staged to [{}]", artifact.name(), artifact.location());
```
honnix, line 150: To avoid confusion that users might think jflyte always stages files no matter they exist at the destination or not.

jflyte/src/main/java/org/flyte/jflyte/ArtifactStager.java
```diff
@@ -77,36 +84,60 @@ static ArtifactStager create(Config config, Collection<ClassLoader> modules) {
...
+      } catch (InterruptedException | ExecutionException e) {
+        if (e instanceof InterruptedException) {
+          Thread.currentThread().interrupt();
+        }
+        for (int j = i; j < futures.size(); ++j) {
+          futures.get(j).cancel(true);
+        }
+
+        throw new RuntimeException(e);
+      }
```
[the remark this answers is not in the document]
honnix: I tried that `getAll` thing but then `stageFiles` became almost empty then I thought it didn't increase readability.

narape (approved):

PR #162, opened 2022-12-20
Auto generate node ID

honnix:
# TL;DR
Auto generate node ID

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
In most cases, there is no need to specify a node ID that must be unique across one workflow. This PR makes it possible to auto-generate one if unspecified.

If node ID is not specified, it will be generated as `w<random>-n0`, `w<random>-n1`, etc. where `random` is a random integer generated per workflow build, and `w<random>` can be specified by a user; and to make a node readable, node name is inferred from class name.

## Tracking Issue
Closes https://github.com/flyteorg/flyte/issues/3185

## Follow-up issue
_NA_

### DIFF
--- a/flytekit-java/src/main/java/org/flyte/flytekit/SdkCondition.java
+++ b/flytekit-java/src/main/java/org/flyte/flytekit/SdkCondition.java
@@ -54,13 +54,6 @@ public class SdkCondition extends SdkTransform {
       List<String> upstreamNodeIds,
       @Nullable SdkNodeMetadata metadata,
       Map<String, SdkBindingData> inputs) {
-    if (metadata != null) {
-      throw new IllegalArgumentException("invariant failed: metadata must be null");
-    }
-    if (!inputs.isEmpty()) {
-      throw new IllegalArgumentException("invariant failed: inputs must be empty");
-    }
-
     SdkBranchNode.Builder nodeBuilder = new SdkBranchNode.Builder(builder);
 
     for (SdkConditionCase case_ : cases) {
--- a/flytekit-java/src/main/java/org/flyte/flytekit/SdkContainerTask.java
+++ b/flytekit-java/src/main/java/org/flyte/flytekit/SdkContainerTask.java
@@ -53,11 +53,6 @@ public abstract class SdkContainerTask<InputT, OutputT> extends SdkTransform
     return "raw-container";
   }
 
-  /** Specifies task name. */
-  public String getName() {
-    return getClass().getName();
-  }
-
   /** Specifies task input type. */
   public SdkType<InputT> getInputType() {
     return inputType;
--- a/flytekit-java/src/main/java/org/flyte/flytekit/SdkNode.java
+++ b/flytekit-java/src/main/java/org/flyte/flytekit/SdkNode.java
@@ -58,6 +58,6 @@ public abstract class SdkNode {
     List<String> upstreamNodeIds =
         getOutputs().isEmpty() ? Collections.singletonList(getNodeId()) : Collections.emptyList();
 
-    return builder.applyInternal(id, transform, upstreamNodeIds, /*metadata=*/ null, getOutputs());
+    return builder.applyInternal(id, transform, upstreamNodeIds, getOutputs());
   }
 }
--- a/flytekit-java/src/main/java/org/flyte/flytekit/SdkPartialTransform.java
+++ b/flytekit-java/src/main/java/org/flyte/flytekit/SdkPartialTransform.java
@@ -98,7 +98,7 @@ class SdkPartialTransform extends SdkTransform {
   }
 
   @Override
-  public SdkTransform withNameOverride(String name, boolean failOnDuplicate) {
+  SdkTransform withNameOverride(String name, boolean failOnDuplicate) {
     requireNonNull(name, "Name override cannot be null");
 
     SdkNodeMetadata newMetadata = SdkNodeMetadata.builder().name(name).build();
--- a/flytekit-java/src/main/java/org/flyte/flytekit/SdkPartialTransform.java
+++ b/flytekit-java/src/main/java/org/flyte/flytekit/SdkPartialTransform.java
@@ -94,10 +94,22 @@ class SdkPartialTransform extends SdkTransform {
 
   @Override
   public SdkTransform withNameOverride(String name) {
+    return withNameOverride(name, true);
+  }
+
+  @Override
+  public SdkTransform withNameOverride(String name, boolean failOnDuplicate) {
     requireNonNull(name, "Name override cannot be null");
 
     SdkNodeMetadata newMetadata = SdkNodeMetadata.builder().name(name).build();
-    checkForDuplicateMetadata(metadata, newMetadata, SdkNodeMetadata::name, "name");
+    try {
+      checkForDuplicateMetadata(metadata, newMetadata, SdkNodeMetadata::name, "name");
+    } catch (IllegalArgumentException e) {
+      if (failOnDuplicate) {
+        throw e;
+      }
+      return this;
+    }
     SdkNodeMetadata mergedMetadata = mergeMetadata(metadata, newMetadata);
 
     return new SdkPartialTransform(transform, fixedInputs, extraUpstreamNodeIds, mergedMetadata);
@@ -112,6 +124,11 @@ class SdkPartialTransform extends SdkTransform {
     SdkNodeMetadata mergedMetadata = mergeMetadata(metadata, newMetadata);
 
     return new SdkPartialTransform(transform, fixedInputs, extraUpstreamNodeIds, mergedMetadata);
+  }
+
+  @Override
+  public String getName() {
+    return transform.getName();
   }
 
   @Override
--- a/flytekit-java/src/main/java/org/flyte/flytekit/SdkPartialTransform.java
+++ b/flytekit-java/src/main/java/org/flyte/flytekit/SdkPartialTransform.java
@@ -94,25 +94,21 @@ class SdkPartialTransform extends SdkTransform {
 
   @Override
   public SdkTransform withNameOverride(String name) {
-    return withNameOverride(name, true);
-  }
-
-  @Override
-  SdkTransform withNameOverride(String name, boolean failOnDuplicate) {
     requireNonNull(name, "Name override cannot be null");
 
     SdkNodeMetadata newMetadata = SdkNodeMetadata.builder().name(name).build();
-    try {
-      checkForDuplicateMetadata(metadata, newMetadata, SdkNodeMetadata::name, "name");
-    } catch (IllegalArgumentException e) {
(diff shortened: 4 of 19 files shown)


flytekit-java/src/main/java/org/flyte/flytekit/SdkCondition.java
```diff
@@ -54,13 +54,6 @@ public SdkBranchNode apply(
       List<String> upstreamNodeIds,
       @Nullable SdkNodeMetadata metadata,
       Map<String, SdkBindingData> inputs) {
-    if (metadata != null) {
```
honnix, line 57: This feels unnecessarily strict.

flytekit-java/src/main/java/org/flyte/flytekit/SdkContainerTask.java
```diff
@@ -53,11 +53,6 @@ public String getType() {
     return "raw-container";
   }
 
-  /** Specifies task name. */
-  public String getName() {
```
honnix, line 57: Moved to `SdkTransform`.

flytekit-java/src/main/java/org/flyte/flytekit/SdkNode.java
```diff
@@ -58,6 +58,6 @@ public SdkNode apply(String id, SdkTransform transform) {
     List<String> upstreamNodeIds =
         getOutputs().isEmpty() ? Collections.singletonList(getNodeId()) : Collections.emptyList();
 
-    return builder.applyInternal(id, transform, upstreamNodeIds, /*metadata=*/ null, getOutputs());
```
honnix, line 61: This is strictly a breaking change, but I'm not sure why we had `applyInternal` as `protected`, and here `metadata` is always `null`.

flytekit-java/src/main/java/org/flyte/flytekit/SdkPartialTransform.java
```diff
@@ -114,6 +126,11 @@ public SdkTransform withTimeoutOverride(Duration timeout) {
     return new SdkPartialTransform(transform, fixedInputs, extraUpstreamNodeIds, mergedMetadata);
   }
 
+  @Override
+  public String getName() {
+    return transform.getName();
```
honnix, line 131: Delegate to `transform` to get real name.

flytekit-java/src/main/java/org/flyte/flytekit/SdkWorkflowBuilder.java
```diff
@@ -47,34 +54,52 @@ public SdkWorkflowBuilder() {
...
-      String nodeId,
+  public SdkNode apply(SdkTransform transform) {
+    return apply(/*nodeId=*/ null, transform, emptyMap());
+  }
+
+  public SdkNode apply(SdkTransform transform, Map<String, SdkBindingData> inputs) {
+    return applyInternal(/*nodeId=*/ null, transform, emptyList(), inputs);
+  }
+
+  SdkNode applyInternal(
```
honnix, line 77: Changed visibility to package only. Note that this is strictly a breaking change.

flytekit-java/src/main/java/org/flyte/flytekit/SdkWorkflowBuilder.java
```diff
@@ -201,4 +226,9 @@ public void output(String name, SdkBindingData value, String help) {
   public WorkflowTemplate toIdlTemplate() {
     return WorkflowTemplateIdl.ofBuilder(this);
   }
+
+  private static String toNodeName(String name) {
+    String lastPart = name.substring(name.lastIndexOf('.') + 1);
+    return PATTERN.matcher(lastPart).replaceAll("$1-$2").toLowerCase().replaceAll("\\$", "-");
```
honnix, line 232: Basically camel to kebab, also placing `$` for nested class.

flytekit-java/src/main/java/org/flyte/flytekit/SdkWorkflowBuilder.java
```diff
@@ -47,34 +55,52 @@ public SdkWorkflowBuilder() {
...
               CompilerError.Kind.DUPLICATE_NODE_ID,
-              nodeId,
+              actualNodeId,
               "Trying to insert two nodes with the same id.");
 
       throw new CompilerException(error);
     }
 
-    SdkNode sdkNode = transform.apply(this, nodeId, upstreamNodeIds, metadata, inputs);
+    String fallbackNodeName =
```
honnix, line 97: For backward compatibility.

flytekit-java/src/main/java/org/flyte/flytekit/SdkCondition.java
honnix: I had to delete these because of calling `SdkTransform withNameOverride(String name, boolean failOnDuplicate)`

flytekit-java/src/main/java/org/flyte/flytekit/SdkWorkflowBuilder.java
```diff
@@ -26,55 +26,89 @@
...
   private final Map<String, SdkNode> nodes;
   private final Map<String, SdkBindingData> inputs;
   private final Map<String, SdkBindingData> outputs;
   private final Map<String, String> inputDescriptions;
   private final Map<String, String> outputDescriptions;
+  private final String nodeIdPrefix;
+  private final AtomicInteger nodeIdSuffix;
 
   public SdkWorkflowBuilder() {
+    this("w" + ThreadLocalRandom.current().nextInt(1000, 10000) + "-");
```
honnix, line 53: Make sure we could support subworkflow later.

flytekit-java/src/main/java/org/flyte/flytekit/SdkTransform.java
```diff
@@ -72,6 +72,10 @@ public SdkTransform withUpstreamNode(SdkNode node) {
   }
 
   public SdkTransform withNameOverride(String name) {
+    return withNameOverride(name, true);
+  }
+
+  SdkTransform withNameOverride(String name, boolean failOnDuplicate) {
```
honnix, line 78: Intentionally not opening this method to avoid exposing unnecessary methods to users.

flytekit-java/src/main/java/org/flyte/flytekit/SdkWorkflowBuilder.java
honnix: I went for consistency because we use number for node already `n0, `n1` etc. This is a also bit simpler than generating random bytes. But I'm ok either way.

flytekit-java/src/main/java/org/flyte/flytekit/SdkTransform.java
```diff
@@ -72,6 +72,10 @@ public SdkTransform withUpstreamNode(SdkNode node) {
   }
 
   public SdkTransform withNameOverride(String name) {
+    return withNameOverride(name, true);
+  }
```
ckiosidis, line 76: this is exposed publicly? So now we have 2 ways to set the name? One though the withNameOverride and the other through the buillder.apply(name,...)?

flytekit-java/src/main/java/org/flyte/flytekit/SdkWorkflowBuilder.java
```diff
@@ -26,55 +26,89 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.regex.Pattern;
 import javax.annotation.Nullable;
 import org.flyte.api.v1.LiteralType;
 import org.flyte.api.v1.SimpleType;
 import org.flyte.api.v1.WorkflowTemplate;
 
 public class SdkWorkflowBuilder {
+
+  private static final Pattern PATTERN = Pattern.compile("([a-z])([A-Z]+)");
```
ckiosidis, line 42: we allow for capitals? i thought that only lower case was allowed

flytekit-java/src/main/java/org/flyte/flytekit/SdkTransform.java
honnix: `buillder.apply(nodeId,...)` is for setting node ID, not node name.

flytekit-java/src/main/java/org/flyte/flytekit/SdkWorkflowBuilder.java
honnix: This is to convert `FooBar` to `foo-bar` for example.

flytekit-java/src/main/java/org/flyte/flytekit/SdkWorkflowBuilder.java
honnix: OK, I revised it to use alphabet as prefix. This will also decrease the chance of conflicting.

ckiosidis (approved):
looks ok to me 👍

rafaelraposospot (commented):
In flytekit-python, it throws exception if not set and there's a method to convert to dns-compliant.

See: 
https://github.com/flyteorg/flytekit/blob/master/flytekit/core/node.py#L28
https://github.com/flyteorg/flytekit/blob/master/flytekit/core/utils.py#L13

Should we align with the Python way? Fine either way for me just want to be sure if we're supposed to be as much aligned as possible or if we can have our own flavor.

honnix:
Where should we throw? In case of an invalid node id? That is reasonable but seems to be better in a different PR.

flytekit-java/src/main/java/org/flyte/flytekit/SdkPartialTransform.java
```diff
@@ -94,10 +94,22 @@ public SdkTransform withUpstreamNode(SdkNode node) {
...
 
     SdkNodeMetadata newMetadata = SdkNodeMetadata.builder().name(name).build();
-    checkForDuplicateMetadata(metadata, newMetadata, SdkNodeMetadata::name, "name");
+    try {
+      checkForDuplicateMetadata(metadata, newMetadata, SdkNodeMetadata::name, "name");
+    } catch (IllegalArgumentException e) {
+      if (failOnDuplicate) {
+        throw e;
+      }
+      return this;
```
narape, line 111: This is surprising for me, if a name is already set and check is false then we don't override the name (this method becomes no op).

My suggestion would be to create a method called `withNameOverrideIfNotSet` (or similar) and then write the method like this:
```java
  public SdkTransform withNameOverrideIfNotSet(String name) {
    if (metadata.name() != null) {
      return this;
    }
    return withNameOverride(name);
  }
```
Leaving the `withNameOverride` as it was original

honnix:
Please feel free to take over and make necessary changes. I will not have time to work on this any time soon. Thank you.

ckiosidis (approved):
looks good

PR #164, opened 2023-01-11
Change SdkBindingData to be typed and add typed transform outputs #minor

sonjaer:
# TL;DR

This PR tries to accomplish two features:

### Change the SdkBindingData to be typed SdkBindingData<?>
This change allows the user to know the inner type of the attributes from INPUT and OUTPUT at development time.

### Add typed transform outputs
This change allows getting the outputs of a SdkNode.getOutputs() this output is a typed output using an AutoValue class or a case class.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

### Change the SdkBindingData to be typed SdkBindingData<?>
This change allows the user to know the inner type of the attributes from INPUT and OUTPUT at development time. Before this change, the users must go to the input/output class to figure out the specified attribute type. Now, the SdkBindingData<?> shows the inner type value making the user experience more smoothie. 

### Add typed transform outputs
This change allows getting the outputs of a SdkNode.getOutputs() this output is a typed output using an AutoValue class or a case class. These changes force all the input/output class attributes to be SdkBindingData<?> and now you need to do a `SdkBindingData.get()` to get the inner value in the run task context ([java example](https://github.com/flyteorg/flytekit-java/blob/05cdccdbaac5df0bface2d7fe17bde95e3ef2790/flytekit-examples/src/main/java/org/flyte/examples/GreetTask.java#L81), [scala example](https://github.com/flyteorg/flytekit-java/blob/05c1a7706f81a392c7774695cad9823bf08c011a/flytekit-examples-scala/src/main/scala/org/flyte/examples/flytekitscala/GreetTask.scala#L47)), but at the same time these changes allow to recover the attributes by name in the workflow site ([java example](https://github.com/flyteorg/flytekit-java/blob/9bd85dce3c7029f65c8b77cb6d34f51913d00702/flytekit-examples/src/main/java/org/flyte/examples/FibonacciWorkflow.java#L42), [scala example](https://github.com/flyteorg/flytekit-java/blob/05c1a7706f81a392c7774695cad9823bf08c011a/flytekit-examples-scala/src/main/scala/org/flyte/examples/flytekitscala/FibonacciWorkflow.scala#L37)).

## Tracking Issue
* https://github.com/flyteorg/flyte/issues/3250
* https://github.com/flyteorg/flyte/issues/3251
## Follow-up issue
* https://github.com/flyteorg/flyte/issues/3252


flytekit-examples-scala/src/main/scala/org/flyte/examples/flytekitscala/AddQuestionTask.scala
```diff
@@ -40,7 +43,9 @@ class AddQuestionTask
     *   the updated greeting message
     */
   override def run(input: AddQuestionTaskInput): AddQuestionTaskOutput =
-    AddQuestionTaskOutput(s"${input.greeting} How are you?")
+    AddQuestionTaskOutput(
+      SdkBindingData.ofString(s"${input.greeting} How are you?")
```
bergman, line 47: Shouldn't this be
```suggestion
      SdkBindingData.ofString(s"${input.greeting.get} How are you?")
```

flytekit-examples-scala/src/main/scala/org/flyte/examples/flytekitscala/DynamicFibonacciWorkflowTask.scala
```diff
@@ -41,21 +44,21 @@ class DynamicFibonacciWorkflowTask
 
     @tailrec
     def fib(
-        n: Long,
-        value: SdkBindingData,
-        prev: SdkBindingData
-    ): SdkBindingData = {
-      if (n == input.n) value
+        n: java.lang.Long,
```
bergman, line 47: Why did this change? Are scala longs not allowed in java generics?

narape: This is something that I am not satisfied, but I have put it aside for now.

In the SdkScalaType there is a comment

>   // TODO we are forced to do this because SdkDataBinding.ofInteger returns a SdkBindingData<java.util.Long>
  //  This makes Scala dev mad when they are forced to use the java types instead of scala types
  //  We need to think what to do, maybe move the factory methods out of SdkDataBinding into their own class
  //  So java and scala can have their own factory class/companion object using their own native types
  //  In the meantime, we need to duplicate all the literal types to use also the java types

We need to find a solution, maybe we can explore this in a new PR and we need to create an assana ticket to track it

flytekit-examples-scala/src/main/scala/org/flyte/examples/flytekitscala/DynamicFibonacciWorkflowTask.scala
```diff
@@ -41,21 +44,21 @@ class DynamicFibonacciWorkflowTask
...
+        value: SdkBindingData[java.lang.Long],
+        prev: SdkBindingData[java.lang.Long]
+    ): SdkBindingData[java.lang.Long] = {
+      if (n == input.n.get()) value
```
sonjaer, line 51:
```suggestion
      if (n == input.n.get) value
```

flytekit-examples-scala/src/main/scala/org/flyte/examples/flytekitscala/DynamicFibonacciWorkflowTask.scala
```diff
@@ -41,21 +44,21 @@ class DynamicFibonacciWorkflowTask
...
           n + 1,
-          builder(s"fib-${n + 1}", SumTask(value, prev)).getOutput("c"),
+          builder(s"fib-${n + 1}", SumTask(value, prev)).getOutputs.c,
           value
         )
     }
 
-    require(input.n > 0, "n < 0")
-    val value = if (input.n == 0) {
+    require(input.n.get() > 0, "n < 0")
```
sonjaer, line 60:
```suggestion
    require(input.n.get > 0, "n < 0")
```

flytekit-examples-scala/src/main/scala/org/flyte/examples/flytekitscala/DynamicFibonacciWorkflowTask.scala
```diff
@@ -41,21 +44,21 @@ class DynamicFibonacciWorkflowTask
...
+    val value = if (input.n.get() == 0) {
```
sonjaer, line 61:
```suggestion
    val value = if (input.n.get == 0) {
```

flytekit-examples-scala/src/main/scala/org/flyte/examples/flytekitscala/GreetTask.scala
```diff
@@ -40,7 +43,7 @@ class GreetTask
     *   the welcome message
     */
   override def run(input: GreetTaskInput): GreetTaskOutput =
-    GreetTaskOutput(s"Welcome, ${input.name}!")
+    GreetTaskOutput(SdkBindingData.ofString(s"Welcome, ${input.name.get()}!"))
```
sonjaer, line 46:
```suggestion
    GreetTaskOutput(SdkBindingData.ofString(s"Welcome, ${input.name.get}!"))
```

flytekit-examples-scala/src/main/scala/org/flyte/examples/flytekitscala/HelloWorldTask.scala
```diff
@@ -16,18 +16,18 @@
...
 class HelloWorldTask
-    extends SdkRunnableTask(
+    extends SdkRunnableTask[HelloWorldTaskInput, Unit](
       SdkScalaType[HelloWorldTaskInput],
       SdkScalaType.unit
     ) {
 
   override def run(input: HelloWorldTaskInput): Unit = {
-    println(input.message)
+    println(input.message.get())
```
sonjaer, line 31:
```suggestion
    println(input.message.get)
```

flytekit-examples-scala/src/main/scala/org/flyte/examples/flytekitscala/HelloWorldTask.scala
```diff
@@ -16,18 +16,18 @@
  */
 package org.flyte.examples.flytekitscala
 
-import org.flyte.flytekit.SdkRunnableTask
+import org.flyte.flytekit.{SdkBindingData, SdkRunnableTask}
 import org.flyte.flytekitscala.SdkScalaType
 
-case class HelloWorldTaskInput(message: String)
+case class HelloWorldTaskInput(message: SdkBindingData[String]) {}
```
sonjaer, line 22:
```suggestion
case class HelloWorldTaskInput(message: SdkBindingData[String]) 
```

flytekit-examples-scala/src/main/scala/org/flyte/examples/flytekitscala/SumTask.scala
```diff
@@ -19,17 +19,21 @@ package org.flyte.examples.flytekitscala
...
 class SumTask
-    extends SdkRunnableTask(
+    extends SdkRunnableTask[SumTaskInput, SumTaskOutput](
       SdkScalaType[SumTaskInput],
       SdkScalaType[SumTaskOutput]
     ) {
 
   override def run(input: SumTaskInput): SumTaskOutput = {
-    SumTaskOutput(input.a + input.b)
+    val result = input.a.get() + input.b.get()
```
sonjaer, line 35:
```suggestion
    val result = input.a.get + input.b.get
```

flytekit-examples-scala/src/main/scala/org/flyte/examples/flytekitscala/SumTask.scala
```diff
@@ -40,6 +44,9 @@ class SumTask
 }
 
 object SumTask {
-  def apply(a: SdkBindingData, b: SdkBindingData): SdkTransform =
+  def apply(
+      a: SdkBindingData[java.lang.Long],
+      b: SdkBindingData[java.lang.Long]
```
sonjaer, line 49:
```suggestion
      a: SdkBindingData[Long],
      b: SdkBindingData[Long]
```

flytekit-examples/src/main/java/org/flyte/examples/AddQuestionTask.java
```diff
@@ -40,7 +40,7 @@ public AddQuestionTask() {
    * @param greeting the input greeting message
    * @return a transformed instance of this class with input data
    */
-  public static SdkTransform of(SdkBindingData greeting) {
+  public static SdkTransform<AddQuestionTask.Output> of(SdkBindingData<?> greeting) {
```
sonjaer, line 43:
```suggestion
  public static SdkTransform<AddQuestionTask.Output> of(SdkBindingData<String> greeting) {
```

flytekit-examples/src/main/java/org/flyte/examples/GreetTask.java
```diff
@@ -36,7 +37,7 @@ public GreetTask() {
    * @param name the input name
    * @return a transformed instance of this class with input data
    */
-  public static SdkTransform of(SdkBindingData name) {
+  public static SdkTransform<Output> of(SdkBindingData<?> name) {
```
sonjaer, line 40:
```suggestion
  public static SdkTransform<Output> of(SdkBindingData<String> name) {
```

flytekit-examples/src/main/java/org/flyte/examples/PhoneBookWorkflow.java
```diff
@@ -37,19 +39,42 @@ public class PhoneBookWorkflow extends SdkWorkflow {
...
+    /**
+     * Wraps the constructor of the generated output value class.
+     *
+     * @param phoneNumbers the String literal output of {@link NodeMetadataExampleWorkflow}
+     * @return output of NodeMetadataExampleWorkflow
+     */
+    public static PhoneBookWorkflow.Output create(List<String> phoneNumbers) {
+      return new AutoValue_PhoneBookWorkflow_Output(
+          SdkBindingData.ofCollection(phoneNumbers, SdkBindingData::ofString));
+    }
```
sonjaer, line 55:
```suggestion
```

Tested removing this and ran the tests locally and it is not used anywhere

flytekit-examples/src/main/java/org/flyte/examples/SimpleStructTask.java
```diff
@@ -32,7 +32,7 @@ public SimpleStructTask() {
     super(JacksonSdkType.of(Input.class), JacksonSdkType.of(Output.class));
   }
 
-  public static SdkTransform of(SdkBindingData struct) {
+  public static SdkTransform<?> of(SdkBindingData<?> struct) {
```
sonjaer, line 35:
```suggestion
  public static SdkTransform<SimpleStructTask.Output> of(SdkBindingData<SimpleStructTask.Input> struct) {
```

sonjaer, line 35: We should add some disclaimer or TODO that we don't support struct tasks yet

flytekit-examples/src/main/java/org/flyte/examples/SumTask.java
```diff
@@ -29,33 +29,39 @@ public SumTask() {
     super(JacksonSdkType.of(SumInput.class), JacksonSdkType.of(SumOutput.class));
   }
 
-  public static SdkTransform of(SdkBindingData a, SdkBindingData b) {
-    return new SumTask().withInput("a", a).withInput("b", b);
+  public static SdkTransform<SumTask.SumOutput> of(SdkBindingData<Long> a, SdkBindingData<Long> b) {
+    return new SumTask()
+        .withInput("a", a) // this still sucks
```
sonjaer, line 34:
```suggestion
        .withInput("a", a)
```

Even though it sucks :D

flytekit-examples/src/main/java/org/flyte/examples/SumTask.java
```diff
@@ -29,33 +29,39 @@ public SumTask() {
...
+        .withInput("b", b);
   }
 
   @AutoValue
   public abstract static class SumInput {
-    public abstract long a();
+    // SdkBindingData<Long> vs long
```
sonjaer, line 40:
```suggestion
```

flytekit-examples/src/main/java/org/flyte/examples/SumTask.java
```diff
@@ -29,33 +29,39 @@ public SumTask() {
...
+    public abstract SdkBindingData<Long> a();
 
-    public abstract long b();
+    public abstract SdkBindingData<Long> b();
 
+    // we haven't decided if this are longs or sdkbinding data too
```
sonjaer, line 45:
```suggestion
```

flytekit-jackson/src/main/java/org/flyte/flytekit/jackson/BindingMapSerializer.java
```diff
@@ -0,0 +1,38 @@
...
+package org.flyte.flytekit.jackson;
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.JsonSerializer;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import java.io.IOException;
+import java.util.Map;
+import org.flyte.flytekit.SdkBindingData;
+
+public class BindingMapSerializer extends JsonSerializer<JacksonBindingMap> {
```
sonjaer, line 26: I would really like we leave some comment on this and other parts somewhere on how this works as I will forget in very short time :P

flytekit-jackson/src/main/java/org/flyte/flytekit/jackson/JacksonSdkType.java
```diff
@@ -42,7 +45,10 @@
       new ObjectMapper()
           .registerModule(new SdkTypeModule())
           .registerModule(new JavaTimeModule())
-          .registerModule(new ParameterNamesModule());
+          .registerModule(new ParameterNamesModule())
+          // TODO: Think about this, this is necessary right now because we are adding literal and
```
sonjaer, line 49: @narape @andresgomezfrr  what is meant by 
```
we are adding literal and scalar inside the JSONode in the case of GENERIC
```?

File: flytekit-jackson/src/main/java/org/flyte/flytekit/jackson/JacksonSdkType.java [hunk 20]
```diff
@@ -123,11 +129,41 @@ public Map<String, Variable> getVariableMap() {
...
+      JsonNode tree = OBJECT_MAPPER.valueToTree(new JacksonLiteralMap(value, literalTypeMap));
 
       return OBJECT_MAPPER.treeToValue(tree, clazz);
     } catch (JsonProcessingException e) {
       throw new RuntimeException("fromLiteralMap failed for [" + clazz.getName() + "]", e);
     }
   }
+
+  @Override
+  public T promiseFor(String nodeId) {
```
sonjaer, line 145: I would really like if we add some comment or doc to this as I will forget. Comment would be nice for now.

flytekit-jackson/src/main/java/org/flyte/flytekit/jackson/JacksonSdkType.java
```diff
@@ -123,11 +129,41 @@ public Map<String, Variable> getVariableMap() {
...
+              .collect(toMap(Map.Entry::getKey, x -> x.getValue().literalType()));
+
+      JsonNode tree = OBJECT_MAPPER.valueToTree(new JacksonLiteralMap(value, literalTypeMap));
 
       return OBJECT_MAPPER.treeToValue(tree, clazz);
     } catch (JsonProcessingException e) {
       throw new RuntimeException("fromLiteralMap failed for [" + clazz.getName() + "]", e);
     }
   }
+
```
sonjaer, line 143:
```suggestion
/**
Method used to create SdkBindingData output references/promises for SdkTransform outputs 
(e.g., workflows and tasks outputs).
We need to go from Map<String, Variable> to object of output class T.
We leverage Jackson to help create the object of the output class T.
We use a the BindingMapSerializer to serialize the keys of the map to JsonNode
Instead of recreating  SdkBindingData objects we pass the bindingMap to the 
CustomSdkBindingDataDeserializers so it can use the serialized keys of the map to retrieve them 
We need to create a new object mapper to use a different deserializer 
for SdkBindingData than the one used in other places. 
*/
```

sonjaer: maybe too much info, will come back to fix this

flytekit-jackson/src/main/java/org/flyte/flytekit/jackson/JacksonSdkType.java
```diff
@@ -123,11 +129,41 @@ public Map<String, Variable> getVariableMap() {
...
+          getVariableMap().entrySet().stream()
+              .collect(
+                  toMap(
+                      Map.Entry::getKey,
+                      x ->
+                          SdkBindingData.ofOutputReference(
+                              nodeId, x.getKey(), x.getValue().literalType())));
+
+      JsonNode tree = OBJECT_MAPPER.valueToTree(new JacksonBindingMap(bindingMap));
+      ObjectMapper mapper =
```
sonjaer, line 157: Move this out to a field so we avoid creating a new object everytime this is called?

flytekit-jackson/src/main/java/org/flyte/flytekit/jackson/LiteralMapSerializer.java
```diff
@@ -0,0 +1,263 @@
...
+      String key,
+      Literal value,
+      JsonGenerator gen,
+      SerializerProvider serializers,
+      Map<String, LiteralType> literalTypeMap)
+      throws IOException {
+
+    switch (value.kind()) {
+      case SCALAR:
+        gen.writeFieldName("literal");
```
sonjaer, line 63: Refactor literal to constant

flytekit-jackson/src/main/java/org/flyte/flytekit/jackson/LiteralMapSerializer.java
```diff
@@ -0,0 +1,263 @@
...
+        gen.writeObject(Literal.Kind.SCALAR);
+        serializeScalar(value.scalar(), gen, serializers, literalTypeMap);
+        return;
+
+      case MAP:
+        gen.writeFieldName("literal");
+        gen.writeObject(Literal.Kind.MAP);
+        gen.writeFieldName("type");
```
sonjaer, line 71: refactor type to constant

flytekit-jackson/src/main/java/org/flyte/flytekit/jackson/LiteralMapSerializer.java
```diff
@@ -0,0 +1,263 @@
...
+        gen.writeObject(literalTypeMap.get(key).mapValueType().simpleType());
+        gen.writeFieldName("value");
```
sonjaer, line 73: refactor value to constant

flytekit-jackson/src/main/java/org/flyte/flytekit/jackson/LiteralMapSerializer.java
```diff
@@ -0,0 +1,263 @@
...
+          gen.writeStartObject();
+          serializeLiteral(entry.getKey(), entry.getValue(), gen, serializers, literalTypeMap);
+          gen.writeEndObject();
+        }
+
+        gen.writeEndObject();
+        return;
+
+      case COLLECTION:
+        gen.writeFieldName("literal");
```
sonjaer, line 87: minor suggestion: we could refactor each case to a function so it is easier to read

flytekit-jackson/src/main/java/org/flyte/flytekit/jackson/LiteralMapSerializer.java
```diff
@@ -0,0 +1,263 @@
...
+        serializePrimitive(value.primitive(), gen);
+        return;
+
+      case GENERIC:
+        gen.writeFieldName("scalar");
+        gen.writeObject(Scalar.Kind.GENERIC);
+        serializeGeneric(value.generic(), gen, serializers, literalTypeMap);
+        return;
+
+      case BLOB:
```
sonjaer, line 126: Should we throw unsupported here?

flytekit-jackson/src/main/java/org/flyte/flytekit/jackson/LiteralMapSerializer.java
```diff
@@ -0,0 +1,263 @@
...
+      Map<String, LiteralType> literalTypeMap)
+      throws IOException {
+    switch (value.kind()) {
+      case PRIMITIVE:
+        gen.writeFieldName("scalar");
+        gen.writeObject(Scalar.Kind.PRIMITIVE);
+        serializePrimitive(value.primitive(), gen);
+        return;
+
+      case GENERIC:
```
sonjaer, line 120: Throw unsupported here and add todo as we don't support it yet in execution so we catch this at deployment time instead

flytekit-jackson/src/main/java/org/flyte/flytekit/jackson/SdkBindingDataDeserializer.java
```diff
@@ -0,0 +1,167 @@
...
+import org.flyte.flytekit.SdkBindingData;
+
+class SdkBindingDataDeserializer extends StdDeserializer<SdkBindingData<?>> {
+  private static final long serialVersionUID = 0L;
+
+  public SdkBindingDataDeserializer() {
+    super(SdkBindingData.class);
+  }
+
+  private SdkBindingData<?> transform(JsonNode tree) {
```
sonjaer, line 48: Refactor this function to avoid nested switch

flytekit-jackson/src/main/java/org/flyte/flytekit/jackson/SdkBindingDataDeserializer.java
```diff
@@ -0,0 +1,167 @@
...
+          case INTEGER:
+            return generateListFromIterator(elements, JsonNode::asLong, SdkBindingData::ofInteger);
+          case FLOAT:
+            return generateListFromIterator(elements, JsonNode::asDouble, SdkBindingData::ofFloat);
+          case BOOLEAN:
+            return generateListFromIterator(
+                elements, JsonNode::asBoolean, SdkBindingData::ofBoolean);
+          case STRUCT:
+            // TODO: need to implement this.
+            throw new RuntimeException("not supported");
```
sonjaer, line 96: Align to throw same exception in all unsupported cases

flytekit-jackson/src/main/java/org/flyte/flytekit/jackson/SdkBindingDataDeserializer.java
```diff
@@ -0,0 +1,167 @@
...
+              case DATETIME:
+                return SdkBindingData.ofDatetime(Instant.parse(tree.get("value").asText()));
+              case FLOAT_VALUE:
+                return SdkBindingData.ofFloat(tree.get("value").doubleValue());
+            }
+            break;
+          case GENERIC:
+          case BLOB:
+            // TODO: We need to implement this
+            throw new RuntimeException("not supported");
```
sonjaer, line 72: Align to throw same exception in all unsupported cases

flytekit-jackson/src/main/java/org/flyte/flytekit/jackson/SdkBindingDataDeserializer.java
sonjaer:
```
            throw new UnsupportedOperationException(
                String.format(
                    "Not supported simple type %s. Literal: %s",
                    collectionSimpleInnerType, literalKind.name()));``

File: flytekit-jackson/src/main/java/org/flyte/flytekit/jackson/SdkBindingDataDeserializer.java [hunk 32]
```diff
@@ -0,0 +1,167 @@
...
+            return SdkBindingData.ofDurationMap(
+                generateMapFromNode(tree, (node) -> Duration.parse(node.asText())));
+          case INTEGER:
+            return SdkBindingData.ofIntegerMap(generateMapFromNode(tree, JsonNode::asLong));
+          case FLOAT:
+            return SdkBindingData.ofFloatMap(generateMapFromNode(tree, JsonNode::asDouble));
+          case BOOLEAN:
+            return SdkBindingData.ofBooleanMap(generateMapFromNode(tree, JsonNode::asBoolean));
+          case STRUCT:
+            // TODO: need to implement this.
```
sonjaer, line 122: Align to throw same exception in all unsupported cases

flytekit-jackson/src/main/java/org/flyte/flytekit/jackson/SdkBindingDataDeserializer.java
```diff
@@ -0,0 +1,167 @@
...
+                    "Not supported simple type %s. Literal: %s",
+                    mapSimpleInnerType, literalKind.name()));
+        }
+
+      default:
+        throw new UnsupportedOperationException(
+            String.format("Not supported literal type %s", literalKind.name()));
+    }
+
+    // TODO: Think about it
```
sonjaer, line 136: Remove default case and move exception here

flytekit-jackson/src/main/java/org/flyte/flytekit/jackson/SdkBindingDataDeserializer.java
```diff
@@ -0,0 +1,167 @@
...
+    super(SdkBindingData.class);
+  }
+
+  private SdkBindingData<?> transform(JsonNode tree) {
+    Literal.Kind literalKind = Literal.Kind.valueOf(tree.get("literal").asText());
+    switch (literalKind) {
+      case SCALAR:
+        switch (Scalar.Kind.valueOf(tree.get("scalar").asText())) {
+          case PRIMITIVE:
+            switch (Primitive.Kind.valueOf(tree.get("primitive").asText())) {
```
sonjaer, line 54: Refactor these to constants

flytekit-jackson/src/test/java/org/flyte/flytekit/jackson/JacksonSdkTypeTest.java
```diff
@@ -174,44 +189,46 @@ public void testPojoVariableMap() {
 
     Map<String, Variable> variableMap = JacksonSdkType.of(PojoInput.class).getVariableMap();
 
-    assertThat(variableMap, equalTo(singletonMap("a", expected)));
+    assertThat(variableMap, equalTo(Map.of("a", expected)));
   }
 
-  @Test
+  @Disabled("Not supported struct with the strongly types implementation.")
```
sonjaer, line 195:
```suggestion
  @Disabled("Not supporting struct with the new strongly typed implementation.")
```

flytekit-jackson/src/test/java/org/flyte/flytekit/jackson/JacksonSdkTypeTest.java
```diff
@@ -174,44 +189,46 @@ public void testPojoVariableMap() {
...
 
     SdkType<StructInput> sdkType = JacksonSdkType.of(StructInput.class);
-
-    assertThat(sdkType.fromLiteralMap(sdkType.toLiteralMap(input)), equalTo(input));
+    Map<String, Literal> literalMap = sdkType.toLiteralMap(input);
+    assertThat(sdkType.fromLiteralMap(literalMap), equalTo(input));
   }
 
-  @Test
+  @Disabled("Not supported customType & customEnum with the strongly types implementation.")
```
sonjaer, line 218:
```suggestion
  @Disabled("Not supporting customType & customEnum with the strongly typed implementation.")
```

flytekit-jackson/src/test/java/org/flyte/flytekit/jackson/JacksonSdkTypeTest.java
```diff
@@ -174,44 +189,46 @@ public void testPojoVariableMap() {
...
 
     Map<String, Literal> literalMap =
-        JacksonSdkType.of(InputWithCustomType.class).toLiteralMap(input);
+         JacksonSdkType.of(InputWithCustomType.class).toLiteralMap(input);
 
     assertThat(literalMap, equalTo(expected));
   }
 
-  @Test
+  @Disabled("Not supported customType & customEnum with the strongly types implementation.")
```
sonjaer, line 231:
```suggestion
  @Disabled("Not supporting customType & customEnum with the strongly typed implementation.")
```

flytekit-jackson/src/test/java/org/flyte/flytekit/jackson/JacksonSdkTypeTest.java
```diff
@@ -224,7 +241,7 @@ public void testConverterFromLiteralMap() {
     assertThat(output, equalTo(expected));
   }
 
-  @Test
+  @Disabled("Not supported customType & customEnum with the strongly types implementation.")
```
sonjaer, line 244:
```suggestion
  @Disabled("Not supporting customType & customEnum with the strongly typed implementation.")
```

flytekit-jackson/src/test/java/org/flyte/flytekit/jackson/JacksonSdkTypeTest.java
```diff
@@ -260,47 +311,72 @@ void testUnknownSerializer() {
...
+        SdkBindingData<List<String>> l,
+        SdkBindingData<Map<String, String>> m) {
+      return new AutoValue_JacksonSdkTypeTest_AutoValueInput(i, f, s, b, t, d, l, m);
     }
   }
 
   @AutoValue
   public abstract static class StructInput {
-    public abstract StructValueInput structValue();
+    public abstract SdkBindingData<StructValueInput> structLevel1();
```
sonjaer, line 377:
```suggestion
    public abstract SdkBindingData<StructValueInput> structLevel();
```

flytekit-jackson/src/test/java/org/flyte/flytekit/jackson/JacksonSdkTypeTest.java
```diff
@@ -309,19 +385,19 @@ public static StructInput create(StructValueInput structValue) {
   public abstract static class StructValueInput {
     public abstract String stringValue();
 
-    public abstract boolean boolValue();
+    public abstract Boolean boolValue();
```
sonjaer, line 388:
```suggestion
    public abstract boolean boolValue();
```

flytekit-jackson/src/test/java/org/flyte/flytekit/jackson/JacksonSdkTypeTest.java
```diff
@@ -309,19 +385,19 @@ public static StructInput create(StructValueInput structValue) {
...
 
-    public abstract List<Integer> listValue();
+    public abstract List<Long> listValue();
```
sonjaer, line 390:
```suggestion
    public abstract List<Integer> listValue();
```

flytekit-jackson/src/test/java/org/flyte/flytekit/jackson/JacksonSdkTypeTest.java
```diff
@@ -309,19 +385,19 @@ public static StructInput create(StructValueInput structValue) {
...
 
     @Nullable
-    public abstract StructValueInput structValue();
+    public abstract StructValueInput structLevel2();
```
sonjaer, line 393:
```suggestion
    public abstract StructValueInput structLevel();
```

flytekit-java/src/main/java/org/flyte/flytekit/CompilerError.java
```diff
@@ -40,7 +40,13 @@ public enum Kind {
...
 
     /** An Id existed twice. */
-    DUPLICATE_NODE_ID
+    DUPLICATE_NODE_ID,
+
+    /** Try to use a typed output without specific a typed output class. */
+    USED_OUTPUT_TRANSFORMER_WITHOUT_SPECIFIC_CLASS,
+
+    /** Some error occurs during the typed output reflection. */
+    OUTPUT_TRANSFORMER_CLASS_REFLECTION_ERROR,
```
sonjaer, line 49:
```suggestion
```
Seems unused

flytekit-java/src/main/java/org/flyte/flytekit/SdkBindingData.java
```diff
@@ -38,215 +37,203 @@
...
-    return ofBindingCollection(elements, SdkBindingData::ofDuration, LiteralTypes.DURATION);
+  public static SdkBindingData<Map<String, Instant>> ofDatetimeMap(Map<String, Instant> map) {
+    return ofMap(map, SdkBindingData::ofDatetime);
   }
 
-  private static <T> SdkBindingData ofBindingCollection(
-      List<T> elements, Function<T, SdkBindingData> f, LiteralType literalType) {
-    List<SdkBindingData> bindings = elements.stream().map(f).collect(toUnmodifiableList());
+  public static <T> SdkBindingData<List<T>> ofBindingCollection(List<SdkBindingData<T>> elements) {
+    // TODO we can fix that by introducing "top type" into type system and
```
sonjaer, line 129: TODO - check if we can fix this in next PR

flytekit-java/src/main/java/org/flyte/flytekit/SdkBindingData.java
```diff
@@ -38,215 +37,203 @@
...
-    return ofBindingMap(elementsByKey, SdkBindingData::ofString, LiteralTypes.STRING);
-  }
-
-  public static SdkBindingData ofBooleanMap(Map<String, Boolean> elementsByKey) {
-    return ofBindingMap(elementsByKey, SdkBindingData::ofBoolean, LiteralTypes.BOOLEAN);
-  }
-
-  public static SdkBindingData ofDatetimeMap(Map<String, Instant> elementsByKey) {
-    return ofBindingMap(elementsByKey, SdkBindingData::ofDatetime, LiteralTypes.DATETIME);
+  public static SdkBindingData<SdkStruct> ofStruct(SdkStruct value) {
```
sonjaer, line 194: We could uncomment this as we don't support structs yet

flytekit-java/src/main/java/org/flyte/flytekit/SdkCondition.java
```diff
@@ -21,42 +21,54 @@
...
 
-  SdkCondition(List<SdkConditionCase> cases, String otherwiseName, SdkTransform otherwise) {
+  SdkCondition(
+      List<SdkConditionCase<OutputT>> cases,
+      String otherwiseName,
+      SdkTransform<OutputT> otherwise) {
     this.cases = cases;
     this.otherwiseName = otherwiseName;
     this.otherwise = otherwise;
+    this.outputType = cases.get(0).then().getOutputType();
```
sonjaer, line 37: Does this work since all have same output type? 
So if you do any conditional all tasks needs to have same output type right?

flytekit-java/src/main/java/org/flyte/flytekit/SdkTransform.java
```diff
@@ -27,62 +27,69 @@
...
+  public SdkTransform<T> withInput(String name, double value) {
     return withInput(name, SdkBindingData.ofFloat(value));
   }
 
-  public SdkTransform withInput(String name, SdkStruct value) {
-    return withInput(name, SdkBindingData.ofStruct(value));
+  // TODO SdkStruct is not strongly typed. We need to decide if we make this a SdkTransform<?>, a
+  // SdkTransform<SdkStruct>
+  //     or even better SdkTransform<T> withInput(String name, SdkType<T> type, T value)
+  public SdkTransform<?> withInput(String name, SdkStruct value) {
```
sonjaer, line 68: comment out as we don't support structs yet

flytekit-java/src/main/java/org/flyte/flytekit/SdkWorkflowBuilder.java
```diff
@@ -97,129 +96,131 @@ SdkNode applyInternal(
...
     return inputOfFloat(name, "");
   }
 
-  public SdkBindingData inputOfFloat(String name, String help) {
+  public SdkBindingData<Float> inputOfFloat(String name, String help) {
     return inputOf(name, LiteralType.ofSimpleType(SimpleType.FLOAT), help);
   }
 
-  public SdkBindingData inputOfStruct(String name) {
+  public SdkBindingData<SdkStruct> inputOfStruct(String name) {
```
sonjaer, line 181: Comment out as we don't support it yet

flytekit-java/src/main/java/org/flyte/flytekit/SdkWorkflowBuilder.java
```diff
@@ -97,129 +96,131 @@ SdkNode applyInternal(
...
     return inputOfStruct(name, "");
   }
 
-  public SdkBindingData inputOfStruct(String name, String help) {
+  public SdkBindingData<SdkStruct> inputOfStruct(String name, String help) {
```
sonjaer, line 185: Comment out as we don't support it yet

flytekit-java/src/test/java/org/flyte/flytekit/SdkWorkflowBuilderTest.java
```diff
@@ -381,15 +381,16 @@ void testUpstreamNode_duplicateWithNode() {
...
   @ParameterizedTest
   @MethodSource("createTransform")
-  void testNodeMetadataOverrides(SdkTransform transform) {
+  void testNodeMetadataOverrides(SdkTransform<TestUnaryIntegerOutput> transform) {
     SdkWorkflowBuilder builder = new SdkWorkflowBuilder();
 
-    SdkBindingData el0 = builder.inputOfInteger("el0");
-    SdkBindingData el1 = builder.inputOfInteger("el1");
+    SdkBindingData<?> el0 = builder.inputOfInteger("el0");
+    SdkBindingData<?> el1 = builder.inputOfInteger("el1");
```
sonjaer, line 388:
```suggestion
    SdkBindingData<Long> el0 = builder.inputOfInteger("el0");
    SdkBindingData<Long> el1 = builder.inputOfInteger("el1");
```

flytekit-java/src/test/java/org/flyte/flytekit/SdkWorkflowBuilderTest.java
```diff
@@ -381,15 +381,16 @@ void testUpstreamNode_duplicateWithNode() {
...
 
-    SdkNode el2 = builder.apply("el2", transform.withInput("a", el0).withInput("b", el1));
+    SdkNode<TestUnaryIntegerOutput> el2 =
+        builder.apply("el2", transform.withInput("a", el0).withInput("b", el1));
 
-    SdkNode el3 =
+    SdkNode<?> el3 =
```
sonjaer, line 393:
```suggestion
    SdkNode<TestUnaryIntegerOutput> el3 =
```

flytekit-java/src/test/java/org/flyte/flytekit/SdkWorkflowBuilderTest.java
```diff
@@ -406,13 +407,14 @@ void testNodeMetadataOverrides(SdkTransform transform) {
 
   @ParameterizedTest
   @MethodSource("createTransform")
-  void testNodeMetadataOverrides_duplicate(SdkTransform transform) {
+  void testNodeMetadataOverrides_duplicate(SdkTransform<TestUnaryIntegerOutput> transform) {
     SdkWorkflowBuilder builder = new SdkWorkflowBuilder();
 
-    SdkBindingData el0 = builder.inputOfInteger("el0");
-    SdkBindingData el1 = builder.inputOfInteger("el1");
+    SdkBindingData<?> el0 = builder.inputOfInteger("el0");
```
sonjaer, line 413:
```suggestion
    SdkBindingData<Long> el0 = builder.inputOfInteger("el0");
```

flytekit-java/src/test/java/org/flyte/flytekit/TestPairIntegerInput.java
```diff
@@ -0,0 +1,64 @@
...
+abstract class TestPairIntegerInput {
+  abstract SdkBindingData<Long> a();
+
+  abstract SdkBindingData<Long> b();
+
+  public static TestPairIntegerInput create(SdkBindingData<Long> a, SdkBindingData<Long> b) {
+    return new AutoValue_TestPairIntegerInput(a, b);
+  }
+
+  public static class SdkType extends org.flyte.flytekit.SdkType<TestPairIntegerInput> {
```
sonjaer, line 34: Remove this? or why is it here?

flytekit-java/src/test/java/org/flyte/flytekit/TestUnaryBooleanOutput.java
```diff
@@ -0,0 +1,58 @@
...
+
+@AutoValue
+abstract class TestUnaryBooleanOutput {
+  abstract SdkBindingData<Boolean> o();
+
+  public static TestUnaryBooleanOutput create(SdkBindingData<Boolean> o) {
+    return new AutoValue_TestUnaryBooleanOutput(o);
+  }
+
+  public static class SdkType extends org.flyte.flytekit.SdkType<TestUnaryBooleanOutput> {
```
sonjaer, line 33: Remove this?

flytekit-scala-tests/src/test/scala/org/flyte/flytekitscala/SdkScalaTypeTest.scala
```diff
@@ -27,13 +26,14 @@ import org.flyte.api.v1.{
...
 }
+import org.flyte.flytekit.SdkBindingData
 import org.junit.Assert.assertEquals
 import org.junit.Test
 
 class SdkScalaTypeTest {
 
   case class ScalarInput(
-      string: String,
+      string: SdkBindingData[String],
```
sonjaer, line 36: SdkBindingData for all other types

flytekit-scala_2.13/src/main/scala/org/flyte/flytekitscala/SdkScalaType.scala
```diff
@@ -113,54 +113,196 @@ object SdkScalaType {
           param.typeclass.fromLiteral(paramLiteral)
         })
       }
+
+      def promiseFor(nodeId: String): T = null.asInstanceOf[T] // TODO
     }
   }
 
-  implicit def stringLiteralType: SdkScalaLiteralType[String] =
+  implicit def sdkBindingLiteralType(implicit
```
sonjaer, line 121: This is still TODO so won't review yet

flytekit-testing/src/main/java/org/flyte/flytekit/testing/SdkTestingExecutor.java
```diff
@@ -59,36 +59,40 @@ public abstract class SdkTestingExecutor {
...
     return SdkTestingExecutor.of(workflow, tasks, workflows);
   }
 
   @Deprecated
-  public static SdkTestingExecutor of(SdkWorkflow workflow, List<SdkRunnableTask<?, ?>> tasks) {
-    ServiceLoader<SdkWorkflow> workflowLoader = ServiceLoader.load(SdkWorkflow.class);
-    List<SdkWorkflow> workflows = new ArrayList<>();
+  public static <T> SdkTestingExecutor of(
+      SdkWorkflow<T> workflow, List<SdkRunnableTask<?, ?>> tasks) {
+    @SuppressWarnings({"unchecked", "rawtypes"})
```
sonjaer, line 85:
```suggestion
```

integration-tests/src/test/java/org/flyte/AdditionalIT.java
```diff
@@ -66,6 +67,7 @@ public void testBranchNodeWorkflow(long a, long b, long c, long d, String expect
     "table-exists,true",
     "non-existent,false",
   })
+  @Disabled
```
sonjaer, line 70:
```suggestion
  @Disabled("Not supporting struct with the strongly typed implementation.")
```

flytekit-java/src/main/java/org/flyte/flytekit/SdkBindingData.java
sonjaer: We'll fix it in the next, added a card in asana

flytekit-jackson/src/main/java/org/flyte/flytekit/jackson/JacksonSdkType.java
sonjaer: Added this as a card in Asana

flytekit-java/src/main/java/org/flyte/flytekit/SdkBindingData.java
```diff
@@ -38,215 +37,203 @@
...
-        return SimpleType.DATETIME;
-      case DURATION:
-        return SimpleType.DURATION;
-    }
-
-    throw new AssertionError("Unexpected Primitive.Kind: " + kind);
+  public static <T> SdkBindingData<List<T>> ofCollection(
+      List<T> collection, Function<T, SdkBindingData<T>> function) {
+    return SdkBindingData.ofBindingCollection(
+        collection.stream().map(function).collect(Collectors.toList()));
```
bergman, line 93: This API feels a little off to me. When is it expected to be used over the (to me) simpler `ofBindingCollection`? I can see for `ofBindingCollection` that one the user needs to wrap their list elements themselves but they likely know well enough how to operate on lists in Java. For `ofCollection` they would need to first understand what the `function` parameter does (could be documented I guess) and then know what `SdkBindingData.of*` function to put there (this they would need to know to wrap the values themselves too though). Perhaps renaming the `function` parameter to `mapper` or `transform` or something could help a bit.

```suggestion
  public static <T> SdkBindingData<List<T>> ofCollection(
      List<T> collection, Function<T, SdkBindingData<T>> mapper) {
    return SdkBindingData.ofBindingCollection(
        collection.stream().map(mapper).collect(Collectors.toList()));
```

bergman: For empty lists, I'm not sure what is worse/best:

```java
SdkBindingData.ofCollection(List.<String>of(), SdkBindingData::ofString);
SdkBindingData.ofBindingCollection(List.<SdkBindingData<String>>of());
```

flytekit-examples-scala/src/main/scala/org/flyte/examples/flytekitscala/SumTask.scala
```diff
@@ -19,17 +19,21 @@ package org.flyte.examples.flytekitscala
 import org.flyte.flytekit.{SdkBindingData, SdkRunnableTask, SdkTransform}
 import org.flyte.flytekitscala.SdkScalaType
 
-case class SumTaskInput(a: Long, b: Long)
-case class SumTaskOutput(c: Long)
+case class SumTaskInput(
+    a: SdkBindingData[java.lang.Long],
+    b: SdkBindingData[java.lang.Long]
```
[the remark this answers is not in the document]
narape: `java.lang.Long` vs scala's `Long` are being tackled on https://github.com/flyteorg/flytekit-java/pull/166, but it needs to converted here so it compiles

flytekit-examples-scala/src/main/scala/org/flyte/examples/flytekitscala/SumTask.scala
narape: See previous comment

flytekit-examples/src/main/java/org/flyte/examples/AllInputsLaunchPlan.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+package org.flyte.examples;
+
+import com.google.auto.service.AutoService;
+import org.flyte.flytekit.SdkLaunchPlan;
+import org.flyte.flytekit.SdkLaunchPlanRegistry;
+import org.flyte.flytekit.SimpleSdkLaunchPlanRegistry;
+
+@AutoService(SdkLaunchPlanRegistry.class)
+public class AllInputsLaunchPlan extends SimpleSdkLaunchPlanRegistry {
```
lupasarin, line 25: Why do we need this new class?

flytekit-examples/src/main/java/org/flyte/examples/SimpleStructTask.java
```diff
@@ -25,14 +24,14 @@
 import org.flyte.flytekit.jackson.JacksonSdkType;
 
 /** Example Flyte task that takes a name as the input and outputs a simple greeting message. */
-@AutoService(SdkRunnableTask.class)
+// @AutoService(SdkRunnableTask.class)
```
lupasarin, line 27: Can we add this back?

flytekit-examples/src/main/java/org/flyte/examples/SumTask.java
lupasarin: So honest! 😆

flytekit-examples/src/main/java/org/flyte/examples/UberWorkflow.java
```diff
@@ -20,25 +20,32 @@
...
+    SdkBindingData<Long> abc =
         builder
             .apply("sub-2", new SubWorkflow().withInput("left", ab).withInput("right", c))
-            .getOutput("result");
-    SdkBindingData abcd = builder.apply("post-sum", SumTask.of(abc, d)).getOutput("c");
-    builder.output("total", abcd);
+            .getOutputs()
+            .result();
+    SdkBindingData<Long> abcd = builder.apply("post-sum", SumTask.of(abc, d)).getOutputs().c();
+    builder.output("result", abcd);
```
lupasarin, line 49: Do we need this line?

flytekit-examples/src/test/java/org/flyte/examples/WorkflowTest.java
```diff
@@ -36,7 +36,7 @@ public void testSubWorkflow() {
             .withFixedInput("d", 4)
             .execute();
 
-    assertEquals(10L, result.getIntegerOutput("total"));
+    assertEquals(10L, result.getIntegerOutput("result"));
```
lupasarin, line 39: shouldn't we use getOutputs().result() instead? (same below)

flytekit-jackson/src/main/java/org/flyte/flytekit/jackson/JacksonSdkType.java
lupasarin: Not sure what we need to do here. @sonjaer Could you please add the link to the Asana ticket?

flytekit-jackson/src/main/java/org/flyte/flytekit/jackson/JacksonSdkType.java
lupasarin: I think it is good to add it for our future understanding of this code 🙂

flytekit-jackson/src/main/java/org/flyte/flytekit/jackson/deserializers/CustomSdkBindingDataDeserializers.java
```diff
@@ -0,0 +1,44 @@
...
+
+import com.fasterxml.jackson.databind.BeanDescription;
+import com.fasterxml.jackson.databind.DeserializationConfig;
+import com.fasterxml.jackson.databind.JavaType;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import com.fasterxml.jackson.databind.deser.Deserializers;
+import java.util.Map;
+import org.flyte.flytekit.SdkBindingData;
+
+// TODO find a better name
```
lupasarin, line 27: Should we do this in this PR?

flytekit-jackson/src/main/java/org/flyte/flytekit/jackson/deserializers/SdkBindingDataDeserializers.java
```diff
@@ -0,0 +1,38 @@
...
+
+  @Override
+  public JsonDeserializer<?> findBeanDeserializer(
+      JavaType type, DeserializationConfig config, BeanDescription beanDesc) {
+    if (SdkBindingData.class.isAssignableFrom(type.getRawClass())) {
+
+      return new SdkBindingDataDeserializer();
+    }
+
+    return null;
```
lupasarin, line 36: does it make sense to throw an exception here instead?

flytekit-jackson/src/main/java/org/flyte/flytekit/jackson/serializers/LiteralMapSerializer.java
```diff
@@ -0,0 +1,58 @@
...
+
+  @Override
+  public void serialize(JacksonLiteralMap map, JsonGenerator gen, SerializerProvider serializers)
+      throws IOException {
+    gen.writeStartObject();
+    map.getLiteralMap().forEach((k, v) -> writeLiteralMapEntry(map, gen, serializers, k, v));
+    gen.writeEndObject();
+  }
+
+  private static void writeLiteralMapEntry(
```
lupasarin, line 41: We use this only because of the lambda. I would use the interface trick like we did in other places of the code (e.g. WriteGenericFunction) with @narape. And remove this, since it is unnecessarily wrapping the IOException into a RuntimeException

flytekit-jackson/src/main/java/org/flyte/flytekit/jackson/serializers/MapSerializer.java
```diff
@@ -0,0 +1,64 @@
...
+    LiteralType valueType = literalType.mapValueType();
+    LiteralTypeSerializer.serialize(valueType, gen);
+    gen.writeFieldName(VALUE);
+    gen.writeStartObject();
+
+    value.map().forEach((k, v) -> writeMapEntry(k, v, valueType));
+    gen.writeEndObject();
+  }
+
+  private void writeMapEntry(String k, Literal v, LiteralType valueType) {
```
lupasarin, line 52: Same regarding using an interface similar to WriteGenericFunction

flytekit-java/src/main/java/org/flyte/flytekit/SdkRemoteLaunchPlan.java
```diff
@@ -43,6 +43,12 @@
 
   public abstract SdkType<OutputT> outputs();
 
+  @Override
+  public SdkType<OutputT> getOutputType() {
+    // TODO consider break backward compatibility to unify the names and avoid this bridge method
```
lupasarin, line 48: Something for the future or for this PR?

flytekit-java/src/main/java/org/flyte/flytekit/SdkTransform.java
lupasarin: throw a meaningful error message instead?

flytekit-java/src/main/java/org/flyte/flytekit/SdkTypes.java
```diff
@@ -41,6 +41,11 @@ public Void fromLiteralMap(Map<String, Literal> value) {
       return null;
     }
 
+    @Override
+    public Void promiseFor(String nodeId) {
+      return null;
```
lupasarin, line 46: Is it correct to return just null here?

flytekit-java/src/main/java/org/flyte/flytekit/SdkBindingData.java
```diff
@@ -38,215 +37,264 @@
...
-        String message =
-            String.format("Type %s doesn't match expected type %s", element.type(), elementType);
+  public static <T> SdkBindingData<List<T>> ofBindingCollection(
+      LiteralType literalType, List<SdkBindingData<T>> elements) {
+    List<BindingData> bindings = elements.stream().map(SdkBindingData::idl).collect(toList());
+    BindingData bindingData = BindingData.ofCollection(bindings);
 
-        throw new IllegalArgumentException(message);
-      }
+    boolean hasPromise = bindings.stream().anyMatch(SdkBindingData::isAPromise);
```
narape, line 214: This method (and other public methods that receive a literal type and SdkBindingData or function to extract one) should check that all the elements are of this same type as the one passed as argument
```suggestion
    List<LiteralType> incompatibleTypes = elements.stream()
            .map(SdkBindingData::type)
            .distinct()
            .filter(type -> !type.equals(literalType))
            .collect(toList());
    if (!incompatibleTypes.isEmpty()) {
      throw new IllegalArgumentException(
              String.format("Type mismatch: expected all elements of type %s but found some elements of type: %s", literalType, incompatibleTypes)
      );
    }

    boolean hasPromise = bindings.stream().anyMatch(SdkBindingData::isAPromise);
```

flytekit-java/src/main/java/org/flyte/flytekit/SdkBindingData.java
```diff
@@ -38,215 +37,264 @@
...
     }
-
-    return ofBindingMap(map, map.values().iterator().next().type());
+    throw new IllegalArgumentException("BindingData.Kind not recognized: " + bindingData.kind());
   }
 
-  public static SdkBindingData ofIntegerMap(Map<String, Long> elementsByKey) {
-    return ofBindingMap(elementsByKey, SdkBindingData::ofInteger, LiteralTypes.INTEGER);
-  }
+  public static <T> SdkBindingData<Map<String, T>> ofBindingMap(
```
narape, line 235: like this one

flytekit-java/src/main/java/org/flyte/flytekit/SdkBindingData.java
narape: Do you mean comment?

flytekit-java/src/main/java/org/flyte/flytekit/SdkCondition.java
narape: Yes, think like a ternary if expression in java or a match in scala that you assign to a variable

flytekit-java/src/main/java/org/flyte/flytekit/SdkTransform.java
```diff
@@ -27,62 +27,69 @@
...
+    return null;
+    // return withInput(name, SdkBindingData.ofStruct(value));
   }
```
narape, line 71:
```suggestion
```

flytekit-java/src/main/java/org/flyte/flytekit/SdkTypes.java
narape: Yes, this `SdkType` is the used for void and `null` if the only acceptable value for a `Void` class

flytekit-java/src/main/java/org/flyte/flytekit/SdkRemoteLaunchPlan.java
narape: For future PR, I created a note for it

flytekit-examples/src/main/java/org/flyte/examples/AllInputsLaunchPlan.java
narape: This PR change type system, this tasks makes a more robust tests as it contains all the types and we can use it to test it live

flytekit-examples/src/main/java/org/flyte/examples/SimpleStructTask.java
narape: Struts are out of scope in this PR

flytekit-examples/src/main/java/org/flyte/examples/UberWorkflow.java
narape: Yes, workflow are kind of soft typed still. We need to call output with a string matching the name of the autovalue

flytekit-jackson/src/main/java/org/flyte/flytekit/jackson/deserializers/CustomSdkBindingDataDeserializers.java
narape: We renamed it but forgot to remove comment

flytekit-examples/src/test/java/org/flyte/examples/WorkflowTest.java
narape: Here this is for testing, we didn't improve that interface on this PR. So it still used the untyped interface for that (attributes as strings)

flytekit-jackson/src/main/java/org/flyte/flytekit/jackson/serializers/LiteralMapSerializer.java
narape: The problem here is that the `forEach` of `Map` requires a `BiConsumer<T,U>`. We cannot change that for our custom interface, therefore the private method is needed to catch the exception

flytekit-jackson/src/main/java/org/flyte/flytekit/jackson/deserializers/SdkBindingDataDeserializers.java
narape: Not here, this is the expected behavior for an implementation of `Deserializers.Base`

flytekit-java/src/test/java/org/flyte/flytekit/TestUnaryBooleanOutput.java
narape: No because I didn\t want to create a circular dependency between the jackson module and this one

narape: No because I didn't want to create a circular dependency between the jackson module and this one

flytekit-jackson/src/main/java/org/flyte/flytekit/jackson/JacksonSdkType.java
narape: The object mapper needs a custom `JacksonBindingMap` on every call

flytekit-jackson/src/main/java/org/flyte/flytekit/jackson/serializers/MapSerializer.java
narape: same response

flytekit-java/src/main/java/org/flyte/flytekit/CompilerError.java
sonjaer: I think it was introduced by Andres. I don't see it here https://github.com/flyteorg/flytekit-java/blob/master/flytekit-java/src/main/java/org/flyte/flytekit/CompilerError.java

flytekit-java/src/main/java/org/flyte/flytekit/SdkBindingData.java
sonjaer: yes

flytekit-java/src/main/java/org/flyte/flytekit/SdkTransform.java
lupasarin: I couldn't find this anymore. I guess we removed it?

flytekit-java/src/main/java/org/flyte/flytekit/SdkWorkflowBuilder.java
lupasarin: Same, couldn't find `inputOfStruct(String name) {` in the code anymore. I guess it's done

flytekit-testing/src/main/java/org/flyte/flytekit/testing/SdkTestingExecutor.java
lupasarin: We still need it. Otherwise we get this warning on IntelliJ: `Unchecked assignment: 'java.util.ServiceLoader' to 'java.util.ServiceLoader<org.flyte.flytekit.SdkWorkflow<?>>'`

flytekit-java/src/main/java/org/flyte/flytekit/SdkBindingData.java
lupasarin: We can also reconsider adding some helper methods for common types like ofStringCollection(...), ofBooleanCollection(...), etc. (but I would say that would be for a future PR)

flytekit-examples/src/main/java/org/flyte/examples/SimpleStructTask.java
lupasarin: I'm wondering if we should leave it commented or just remove it. It also depends on how close in the future we will work on the structs implementation. But I'm fine with both leaving it commented or removing it.

flytekit-java/src/main/java/org/flyte/flytekit/SdkRemoteLaunchPlan.java
lupasarin: Cool, thanks! Where are you collecting your notes BTW?

flytekit-java/src/main/java/org/flyte/flytekit/CompilerError.java
andresgomezfrr: Yeah! it was introduced by me on the original PR. We can remove it.

andresgomezfrr (commented):
Reviewed!

flytekit-java/src/main/java/org/flyte/flytekit/SdkBindingData.java
andresgomezfrr: @pablocasares added these helper methods.

PR #169, opened 2023-01-17
Add support for empty Lists and empty Maps

pablocasares:
# TL;DR
Add support for empty Lists and empty Maps in SdkBindingData. 
Add helper methods to create collections of most common types:
- ofStringCollection
- ofFloatCollection
- ofIntegerCollection
- ofBooleanCollection
- ofDurationCollection
- ofDatetimeCollection

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
Added new helpers in SdkBindingData class and modified methods ofMap and ofCollection to ask the user to pass the LiteralType in order to be able to know the underlying type of the collection/map.

### DIFF
--- a/flytekit-examples/src/main/java/org/flyte/examples/PhoneBookWorkflow.java
+++ b/flytekit-examples/src/main/java/org/flyte/examples/PhoneBookWorkflow.java
@@ -51,7 +51,7 @@ public class PhoneBookWorkflow extends SdkWorkflow<PhoneBookWorkflow.Output> {
      */
     public static PhoneBookWorkflow.Output create(List<String> phoneNumbers) {
       return new AutoValue_PhoneBookWorkflow_Output(
-          SdkBindingData.ofCollection(phoneNumbers, SdkBindingData::ofString));
+          SdkBindingData.ofStringCollection(phoneNumbers));
     }
   }
 
@@ -63,8 +63,7 @@ public class PhoneBookWorkflow extends SdkWorkflow<PhoneBookWorkflow.Output> {
   public void expand(SdkWorkflowBuilder builder) {
     SdkBindingData<Map<String, String>> phoneBook = SdkBindingData.ofStringMap(PHONE_BOOK);
 
-    SdkBindingData<List<String>> searchKeys =
-        SdkBindingData.ofCollection(NAMES, SdkBindingData::ofString);
+    SdkBindingData<List<String>> searchKeys = SdkBindingData.ofStringCollection(NAMES);
 
     SdkBindingData<List<String>> phoneNumbers =
         builder
--- a/flytekit-java/src/main/java/org/flyte/flytekit/SdkBindingData.java
+++ b/flytekit-java/src/main/java/org/flyte/flytekit/SdkBindingData.java
@@ -88,63 +88,134 @@ public abstract class SdkBindingData<T> {
   }
 
   public static <T> SdkBindingData<List<T>> ofCollection(
-      List<T> collection, Function<T, SdkBindingData<T>> function) {
+      List<T> collection, LiteralType literalType, Function<T, SdkBindingData<T>> function) {
     return SdkBindingData.ofBindingCollection(
-        collection.stream().map(function).collect(Collectors.toList()));
+        literalType, collection.stream().map(function).collect(Collectors.toList()));
+  }
+
+  private static <T> SdkBindingData<List<T>> createCollection(
+      List<T> collection, LiteralType literalType, Function<T, BindingData> bindingDataFn) {
+    return create(
+        BindingData.ofCollection(
+            collection.stream().map(bindingDataFn).collect(Collectors.toList())),
+        literalType,
+        collection);
+  }
+
+  public static SdkBindingData<List<String>> ofStringCollection(List<String> collection) {
+    return createCollection(
+        collection,
+        LiteralType.ofCollectionType(LiteralType.ofSimpleType(SimpleType.STRING)),
+        (value) -> BindingData.ofScalar(Scalar.ofPrimitive(Primitive.ofStringValue(value))));
+  }
+
+  public static SdkBindingData<List<Double>> ofFloatCollection(List<Double> collection) {
+    return createCollection(
+        collection,
+        LiteralType.ofCollectionType(LiteralType.ofSimpleType(SimpleType.FLOAT)),
+        (value) -> BindingData.ofScalar(Scalar.ofPrimitive(Primitive.ofFloatValue(value))));
+  }
+
+  public static SdkBindingData<List<Long>> ofIntegerCollection(List<Long> collection) {
+    return createCollection(
+        collection,
+        LiteralType.ofCollectionType(LiteralType.ofSimpleType(SimpleType.INTEGER)),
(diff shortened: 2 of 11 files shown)


andresgomezfrr (approved):

flytekit-examples/src/main/java/org/flyte/examples/PhoneBookWorkflow.java
```diff
@@ -51,7 +51,7 @@ public abstract static class Output {
      */
     public static PhoneBookWorkflow.Output create(List<String> phoneNumbers) {
       return new AutoValue_PhoneBookWorkflow_Output(
-          SdkBindingData.ofCollection(phoneNumbers, SdkBindingData::ofString));
+          SdkBindingData.ofStringCollection(phoneNumbers));
```
narape, line 54: Nice that you retrofitted the new methods on old calls

flytekit-java/src/main/java/org/flyte/flytekit/SdkBindingData.java
```diff
@@ -88,63 +88,134 @@ private static <T> SdkBindingData<T> ofPrimitive(Function<T, Primitive> toPrimit
...
   public static <T> SdkBindingData<List<T>> ofCollection(
-      List<T> collection, Function<T, SdkBindingData<T>> function) {
+      List<T> collection, LiteralType literalType, Function<T, SdkBindingData<T>> function) {
     return SdkBindingData.ofBindingCollection(
-        collection.stream().map(function).collect(Collectors.toList()));
+        literalType, collection.stream().map(function).collect(Collectors.toList()));
+  }
+
+  private static <T> SdkBindingData<List<T>> ofTCollection(
+      List<T> collection, LiteralType literalType, Function<T, BindingData> bindingDataFn) {
```
narape, line 97:
```suggestion
  interface ToBindingDataFunction<T> {
    BindingData apply(T value);
  }
      List<T> collection, LiteralType literalType, ToBindingDataFunction bindingDataFn) {
```
```suggestion
  private static <T> SdkBindingData<List<T>> ofTCollection(
```
You had to change the name of the method because its eraused firm is the same as the other. But you gave it a name that it is applicable to both methods, both receive a `List<T>`

I think that if we introduce our own interface we can break the erauser problem and keep using the `ofCollection` name

flytekit-java/src/main/java/org/flyte/flytekit/SdkBindingData.java
```diff
@@ -88,63 +88,134 @@ private static <T> SdkBindingData<T> ofPrimitive(Function<T, Primitive> toPrimit
...
     return SdkBindingData.ofBindingMap(
+        literalType,
         map.entrySet().stream()
             .map(e -> Map.entry(e.getKey(), bindingFunction.apply(e.getValue())))
             .collect(toMap(Map.Entry::getKey, Map.Entry::getValue)));
   }
 
-  public static SdkBindingData<Map<String, Double>> ofFloatMap(Map<String, Double> map) {
-    return ofMap(map, SdkBindingData::ofFloat);
+  private static <T> SdkBindingData<Map<String, T>> ofTMap(
```
narape, line 156: Same comment here, we can reuse the same interface

flytekit-java/src/test/java/org/flyte/flytekit/SdkBindingDataTest.java
```diff
@@ -60,15 +65,126 @@ public void testOfBindingCollection() {
 
   @Test
   public void testOfBindingCollection_empty() {
-    IllegalArgumentException e =
-        assertThrows(
-            IllegalArgumentException.class, () -> SdkBindingData.ofBindingCollection(emptyList()));
+    List<BindingData> expected = emptyList();
```
narape, line 68: Nice

flytekit-java/src/test/java/org/flyte/flytekit/SdkBindingDataTest.java
```diff
@@ -60,15 +65,126 @@ public void testOfBindingCollection() {
...
+    SdkBindingData<List<Long>> output =
+        SdkBindingData.create(
+            BindingData.ofCollection(expected),
+            LiteralType.ofCollectionType(LiteralTypes.INTEGER),
+            emptyList());
 
     assertThat(
-        e.getMessage(),
+        output,
```
narape, line 76: These assertions are upside down, we assume that the create method from autovalue works correctly so what we really want to test here is the 
```java
            SdkBindingData.ofBindingCollection(
                LiteralType.ofCollectionType(LiteralType.ofSimpleType(SimpleType.INTEGER)),
                emptyList())));
```
So that is what should be as the first parameter of the `assertThat`
So 
* rename `output` to `expected`
* what you have inside of the `equalsTo` name it as `output` before the whole `assertThat`
* Then the assertThat should be 
```java
    assertThat(output, equalTo(expected));
```

narape: The same applies to other tests

flytekit-java/src/main/java/org/flyte/flytekit/SdkBindingData.java
```diff
@@ -153,8 +153,8 @@ public static <T> SdkBindingData<Map<String, T>> ofMap(
             .collect(toMap(Map.Entry::getKey, Map.Entry::getValue)));
   }
 
-  private static <T> SdkBindingData<Map<String, T>> ofTMap(
-      Map<String, T> map, LiteralType literalType, Function<T, BindingData> bindingDataFn) {
+  private static <T> SdkBindingData<Map<String, T>> ofMapFromBindingData(
```
narape, line 156: Name it `createMap` here

flytekit-java/src/test/java/org/flyte/flytekit/SdkBindingDataTest.java
```diff
@@ -86,15 +84,16 @@ public void testOfStringCollection() {
     input.add(SdkBindingData.ofString("1"));
     input.add(SdkBindingData.ofString("2"));
 
-    List<String> expected = new ArrayList<>();
-    expected.add("1");
-    expected.add("2");
+    List<String> expectedValue = new ArrayList<>();
+    expectedValue.add("1");
+    expectedValue.add("2");
```
narape, line 89:
```suggestion
    List<String> expectedValue = List.of("1", "2");
```
I think that we are in java 11, so we can use this now

narape: here and the rests of the tests

flytekit-java/src/test/java/org/flyte/flytekit/SdkBindingDataTest.java
```diff
@@ -233,14 +239,15 @@ public void testOfStringMap() {
     input.put("a", SdkBindingData.ofString("1"));
     input.put("b", SdkBindingData.ofString("2"));
 
-    Map<String, String> expected = new HashMap<>();
-    expected.put("a", "1");
-    expected.put("b", "2");
+    Map<String, String> expectedValue = new HashMap<>();
```
narape, line 242: We can use `Map.of` instead of the `HashMap`s

PR #186, opened 2023-02-02
Add JavaDoc comments to flytekit-api module

pablocasares:
Add JavaDoc comments to flytekit-api module following the protos in flyteidl-protos  module

## Type
 - [x] Documentation
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

### DIFF
--- a/flytekit-api/src/main/java/org/flyte/api/v1/Container.java
+++ b/flytekit-api/src/main/java/org/flyte/api/v1/Container.java
@@ -24,16 +24,31 @@ import javax.annotation.Nullable;
 @AutoValue
 public abstract class Container {
 
+  /**
+   * Command to be executed, if not provided, the default entrypoint in the container image will be
+   * used.
+   */
   public abstract List<String> command();
 
+  /**
+   * These will default to Flyte given paths. If provided, the system will not append known paths.
+   * If the task still needs flyte's inputs and outputs path, add $(FLYTE_INPUT_FILE),
+   * $(FLYTE_OUTPUT_FILE) wherever makes sense and the system will populate these before executing
+   * the container.
+   */
   public abstract List<String> args();
 
+  /** Container image url. Eg: docker/redis:latest */
   public abstract String image();
 
+  /** Environment variables will be set as the container is starting up. */
   public abstract List<KeyValuePair> env();
 
+  /** Container resources requirement as specified by the container engine. */
   @Nullable
   public abstract Resources resources();
+
+  // TODO: add ports and architecture from src/main/proto/flyteidl/core/tasks.proto
 
   public static Builder builder() {
     return new AutoValue_Container.Builder();
--- a/flytekit-api/src/main/java/org/flyte/api/v1/ContainerError.java
+++ b/flytekit-api/src/main/java/org/flyte/api/v1/ContainerError.java
@@ -16,11 +16,14 @@
  */
 package org.flyte.api.v1;
 
-/** Exception to propagate detailed errors from container to the execution. */
+/** Error message to propagate detailed errors from container executions to the execution engine. */
 public class ContainerError extends RuntimeException {
 
   private static final long serialVersionUID = 5162780469952221158L;
+  /** A simplified code for errors, so that we can provide a glossary of all possible errors. */
   private final String code;
+  // TODO: add message from src/main/proto/flyteidl/core/errors.proto
+  /** An abstract error kind for this error. Defaults to Non_Recoverable if not specified. */
   private final Kind kind;
 
   /** Defines a generic error type that dictates the behavior of the retry strategy. */
--- a/flytekit-api/src/main/java/org/flyte/api/v1/DynamicJobSpec.java
+++ b/flytekit-api/src/main/java/org/flyte/api/v1/DynamicJobSpec.java
@@ -20,6 +20,7 @@ import com.google.auto.value.AutoValue;
 import java.util.List;
 import java.util.Map;
 
+/** Describes a set of tasks to execute and how the final outputs are produced. */
 @AutoValue
 public abstract class DynamicJobSpec {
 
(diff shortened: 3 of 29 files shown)


flytekit-api/src/main/java/org/flyte/api/v1/DynamicJobSpec.java
```diff
@@ -39,14 +42,14 @@ public abstract class DynamicJobSpec {
   public abstract List<Binding> outputs();
 
   /**
-   * Returns sub-workflows templates.
+   * [Optional] A complete list of task specs referenced in nodes.
```
narape, line 45: Can you check here, this doesnät looks like it belongs to subworkflows

pablocasares: Hi @narape, I checked both the proto from flyteidl-protos and the original proto from [here](https://github.com/flyteorg/flyteidl/blob/master/gen/pb-go/flyteidl/core/dynamic_job.pb.go#L38) and it seems that the comment in flytekit-java is following the one in flyteidl repo. Do you think we should change it to something different?

narape: It looks like a copy paste error (in the proto) of the field bellow. Keep the comment that it was before

flytekit-api/src/main/java/org/flyte/api/v1/Container.java
```diff
@@ -24,17 +24,32 @@
...
   public abstract String image();
 
+  /** Environment variables will be set as the container is starting up. */
   public abstract List<KeyValuePair> env();
 
+  /** Container resources requirement as specified by the container engine. */
   @Nullable
   public abstract Resources resources();
 
+  // TODO: add ports and architecture from src/main/proto/flyteidl/core/tasks.proto
```
lupasarin, line 51: I guess to be resolved in a separate PR? Just checking if it's something we forgot

flytekit-api/src/main/java/org/flyte/api/v1/ContainerError.java
```diff
@@ -16,11 +16,14 @@
  */
 package org.flyte.api.v1;
 
-/** Exception to propagate detailed errors from container to the execution. */
+/** Error message to propagate detailed errors from container executions to the execution engine. */
 public class ContainerError extends RuntimeException {
 
   private static final long serialVersionUID = 5162780469952221158L;
+  /** A simplified code for errors, so that we can provide a glossary of all possible errors. */
```
lupasarin, line 23: Are there known strings that we know the full list of them? If so, can we add a link to that list?

flytekit-api/src/main/java/org/flyte/api/v1/ContainerError.java
```diff
@@ -16,11 +16,14 @@
...
   private final String code;
+  // TODO: add message from src/main/proto/flyteidl/core/errors.proto
```
lupasarin, line 25: Is this TODO to be solved here?

flytekit-api/src/main/java/org/flyte/api/v1/LaunchPlanIdentifier.java
```diff
@@ -28,13 +28,19 @@ public static Builder builder() {
 
   @AutoValue.Builder
   public abstract static class Builder {
-
+    /**
+     * Name of the domain the resource belongs to. A domain can be considered as a subset within a
```
lupasarin, line 32:
```suggestion
     * Name of the domain the launch plan belongs to. A domain can be considered as a subset within a
```
(same below)

flytekit-api/src/main/java/org/flyte/api/v1/NamedEntityIdentifier.java
```diff
@@ -19,15 +19,26 @@
...
+   * specific project.
+   */
   public abstract String domain();
 
+  /** Name of the project the resource belongs to. */
   public abstract String project();
 
+  /**
+   * User provided value for the resource. The combination of project + domain + name uniquely
+   * identifies the resource. [Optional] - in certain contexts - like 'List API', 'Launch plans'
```
lupasarin, line 40:
```suggestion
   * identifies the resource. [Optional] - used in certain contexts - like 'List API', 'Launch plans'
```

narape (approved):

flytekit-api/src/main/java/org/flyte/api/v1/TaskTemplate.java
```diff
@@ -29,19 +29,39 @@ public abstract class TaskTemplate {
...
+  /**
+   * A strongly typed interface for the task. This enables others to use this task within a workflow
+   * and guarantees compile-time validation of the workflow to avoid costly runtime failures.
+   */
   public abstract TypedInterface interface_();
 
+  /**
+   * A predefined yet extensible Task type identifier. This can be used to customize any of the
+   * components. If no extensions are provided in the system, Flyte will resolve this task to its
+   * TaskCategory and default the implementation registered for the TaskCategory.
```
lupasarin, line 41: I don't fully understand this part: "If no extensions are provided in the system, Flyte will resolve this task to its
   * TaskCategory and default the implementation registered for the TaskCategory."

flytekit-api/src/main/java/org/flyte/api/v1/Variable.java
```diff
@@ -23,8 +23,10 @@
 @AutoValue
 public abstract class Variable {
 
+  /** Variable literal type. */
   public abstract LiteralType literalType();
 
+  /** [Optional] string describing input variable. */
```
lupasarin, line 29: Is Variable always used to refer to an input?

flytekit-api/src/main/java/org/flyte/api/v1/Container.java
pablocasares: The added TODO's should not be resolved in this PR. They were added to track missing fields in the java classes from the proto files.

PR #196, opened 2023-02-20
Do not delegate class loading to parent for some classes

honnix:
# TL;DR
Only load certain classes from child.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
Unfortunately https://github.com/flyteorg/flytekit-java/pull/145 did not fully fix https://github.com/flyteorg/flyte/issues/3147 because it only fixed resource, not class.

This PR will skip delegating to parent for class load as well.

This is what slf4j <2.0.x does:

```java
            if (!isAndroid()) {
                staticLoggerBinderPathSet = findPossibleStaticLoggerBinderPathSet();
                reportMultipleBindingAmbiguity(staticLoggerBinderPathSet);
            }
            // the next line does the binding
            StaticLoggerBinder.getSingleton();
```
`findPossibleStaticLoggerBinderPathSet` does the resource load so even we prevent delegation to parent at that point, `StaticLoggerBinder.getSingleton()` will still delegate to parent, without this PR.

## Tracking Issue

fixes https://github.com/flyteorg/flyte/issues/3147

## Follow-up issue
_NA_

### DIFF
--- a/jflyte/src/main/java/org/flyte/jflyte/ChildFirstClassLoader.java
+++ b/jflyte/src/main/java/org/flyte/jflyte/ChildFirstClassLoader.java
@@ -25,8 +25,6 @@ import java.util.Enumeration;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 /**
  * {@link URLClassLoader} that loads classes into child class loader, instead of parent.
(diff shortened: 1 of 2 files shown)


jflyte/src/main/java/org/flyte/jflyte/ChildFirstClassLoader.java
```diff
@@ -134,12 +129,20 @@ public Enumeration<URL> getResources(String name) throws IOException {
...
+
+  private static boolean startsWith(String name, Set<String> prefixes) {
+    for (String resourcePrefix : prefixes) {
       if (name.startsWith(resourcePrefix)) {
-        return false;
+        return true;
       }
     }
-    return true;
+    return false;
```
narape, line 146:
```java
    return prefixes.stream().anyMatch(prefix -> name.startsWith(prefix));
```

narape (commented):
Made a small comment

jflyte/src/main/java/org/flyte/jflyte/ChildFirstClassLoader.java
honnix: I'm not sure which style is easier to read, but I'm fine with this.

honnix: There are probably hundreds of lines of code behind this `anyMatch`, lol.

narape (approved):

PR #198, opened 2023-02-27
Literaltype as sdktype

narape:
# TL;DR
Use  SdkLiteralType to create SdkType<T> with low overhead.

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [ ] Smoke tested
 - [X] Unit tests added
 - [X] Code documentation added

## Complete description
Currently the way to create custom `SdkType<T>` in Java is using `JacksonSdkType`. For example

If a developer wants to define a task equivalent to (String str) -> str.toUpperCase() currently he/she needs to write it as

```java
@AutoService(SdkRunnableTask.class)
public class ToUpperCaseTask extends SdkRunnableTask<ToUpperCaseTask.Input, ToUpperCaseTask.Output> {
  public ToUpperCaseTask() {
    super(JacksonSdkType.of(Input.class), JacksonSdkType.of(Output.class));
  }

  @AutoValue
  public abstract static class Input {
    public abstract SdkBindingData<String> in();
  }

  @AutoValue
  public abstract static class Output {
    public abstract SdkBindingData<String> out();

    public static Output create(String out) {
      return new AutoValue_ToUpperCaseTask_Output(out);
    }
  }

  @Override
  public Output run(Input input) {
    return Output.create(SdkBindingDataFactory.of(input.get().toUpperCase()));
  }
}
```
For such trivial task, defining the AutoValues to define the types takes the majority of the lines.

With this PR the developer could write the task this way instead

```java
@AutoService(SdkRunnableTask.class)
public class ToUpperCaseTask extends SdkRunnableTask<SdkBindingData<String>, SdkBindingData<String>> {
  public ToUpperCaseTask() {
    super(SdkTypes.of(SdkLiteralTypes.strings(), "str"), SdkTypes.of(SdkLiteralTypes.strings(), "upr"));
  }

  @Override
  public SdkBindingData<String> run(SdkBindingData<String> input) {
    return input.get().toUpperCase();
  }
}
```

Defining the types for the task now takes one line only and the run is simplified too as no need to create or access AutoValues

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/3378

## Follow-up issue
_NA_

### DIFF
--- a/flytekit-java/src/main/java/org/flyte/flytekit/MonoVariableSdkType.java
+++ b/flytekit-java/src/main/java/org/flyte/flytekit/MonoVariableSdkType.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright 2021 Flyte Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.flyte.flytekit;
+
+import static java.util.Objects.requireNonNull;
+
+import java.util.Map;
+import org.flyte.api.v1.Literal;
+import org.flyte.api.v1.Variable;
+
+/**
+ * Implementation of {@link SdkType} for types that only contain a single variable.
+ *
+ * @param <T> Native type for the type of the single variable.
+ */
+class MonoVariableSdkType<T> extends SdkType<SdkBindingData<T>> {
+
+  private final SdkLiteralType<T> sdkLiteralType;
+  private final String varName;
+  private final String varDescription;
+
+  MonoVariableSdkType(SdkLiteralType<T> sdkLiteralType, String varName, String varDescription) {
+    this.sdkLiteralType = requireNonNull(sdkLiteralType, "the sdkLiteralType should not be null");
+    this.varName = requireNonNull(varName, "the variable name cannot be null");
+    this.varDescription = requireNonNull(varDescription, "the description cannot be null");
+  }
+
+  @Override
+  public Map<String, Literal> toLiteralMap(SdkBindingData<T> value) {
+    return Map.of(varName, sdkLiteralType.toLiteral(value.get()));
+  }
+
+  @Override
+  public SdkBindingData<T> fromLiteralMap(Map<String, Literal> literalMap) {
+    if (!literalMap.containsKey(varName)) {
+      throw new IllegalArgumentException(
+          String.format("variable %s not found in literal map", varName));
+    }
+    return SdkBindingData.literal(
+        sdkLiteralType, sdkLiteralType.fromLiteral(literalMap.get(varName)));
+  }
+
+  @Override
+  public SdkBindingData<T> promiseFor(String nodeId) {
+    return SdkBindingData.promise(sdkLiteralType, nodeId, varName);
+  }
+
+  @Override
+  public Map<String, Variable> getVariableMap() {
+    return Map.of(
+        varName,
+        Variable.builder()
+            .literalType(sdkLiteralType.getLiteralType())
+            .description(varDescription)
+            .build());
+  }
+
+  @Override
+  public Map<String, SdkLiteralType<?>> toLiteralTypes() {
+    return Map.of(varName, sdkLiteralType);
+  }
+
+  @Override
+  public Map<String, SdkBindingData<?>> toSdkBindingMap(SdkBindingData<T> value) {
+    return Map.of(varName, value);
+  }
+}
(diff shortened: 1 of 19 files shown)


flytekit-java/src/main/java/org/flyte/flytekit/MonoVariableSdkType.java
```diff
@@ -0,0 +1,64 @@
...
+
+import java.util.Map;
+import org.flyte.api.v1.Literal;
+import org.flyte.api.v1.Variable;
+
+/**
+ * Implementation of {@link SdkType} for types that only contain a single variable.
+ * @param <T> Native type for the type of the single variable.
+ */
+class MonoVariableSdkType<T> extends SdkType<SdkBindingData<T>> {
```
narape, line 13: **Note to reviewers**
This class is package private but if you have a better name please say so

flytekit-java/src/test/java/org/flyte/flytekit/SdkNodeNamePolicyTest.java
```diff
@@ -62,7 +62,7 @@ void nextNodeIdShouldReturnSamePrefixForSamePolicy() {
   void nextNodeIdShouldReturnUniquePrefixForDistinctPolicies() {
     var prefixes =
         Stream.generate(SdkNodeNamePolicy::new)
-            .limit(100)
+            .limit(10)
```
narape, line 65: This is unrelated to this change but when creating this PR this test failed.
It is because the birth paradox, I reduced the number of samples to reduce the probability to a really small one

flytekit-scala_2.13/src/main/scala/org/flyte/flytekitscala/SdkScalaType.scala
```diff
@@ -198,21 +198,6 @@ object SdkScalaType {
       sdkLiteral: SdkScalaLiteralType[T]
   ): SdkScalaLiteralType[SdkBindingData[T]] = {
 
-    def toBindingData(literal: Literal): BindingData = {
```
narape, line 201: This piece of code was already unused before the changes of this PR

flytekit-scala_2.13/src/main/scala/org/flyte/flytekitscala/SdkScalaType.scala
```diff
@@ -279,8 +264,33 @@ object SdkScalaType {
 
   def apply[T <: Product]: SdkScalaProductType[T] = macro Magnolia.gen[T]
 
+  /** Returns a [[SdkType]] for [[Unit]] which contains no properties.
```
narape, line 267: We missed this comment as well in a previous PR

andresgomezfrr:
@narape How this new type of definition looks with the strongly types? Could you show an example of usage in a workflow?

andresgomezfrr (commented):
@narape How this new type of definition looks with the strongly types? Could you show an example of usage in a workflow?

narape:
@andresgomezfrr I guess your question comes from passing the name of the property as string, but take a look at the SumTask and now with the latest commit to the FibbonacciWorkflow. 
For example
```scala
    val fib2: SdkBindingData[Long] = builder
      .apply("fib-2", new SumTask(), SumTaskInput(input.fib0, input.fib1))
      .getOutputs
```
notice now `getOutputs` return the `SdkBindingData[Long]` directly, no need to call any `.c()` method. This is why the name of the variable could be defined as string, because it would be hidden from the wiring code, yet still present internally in the promises of the idl

flytekit-java/src/main/java/org/flyte/flytekit/MonoVariableSdkType.java
bergman: I think the [arity](https://en.wikipedia.org/wiki/Arity) concept matches this quite well, so `nullary` (0), `unary` (1), `binary` (2), …, `variadic`(n) are the common prefixes for those.
```suggestion
class UnaryVariableSdkType<T> extends SdkType<SdkBindingData<T>> {
```

flytekit-java/src/main/java/org/flyte/flytekit/MonoVariableSdkType.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.Map;
+import org.flyte.api.v1.Literal;
+import org.flyte.api.v1.Variable;
+
+/**
+ * Implementation of {@link SdkType} for types that only contain a single variable.
+ *
+ * @param <T> Native type for the type of the single variable.
+ */
+class MonoVariableSdkType<T> extends SdkType<SdkBindingData<T>> {
```
rafaelraposospot, line 30: I would recommend to change the `Mono...` it's correct in Spanish/Portuguese but it's not a thing in the English languaage. Maybe switch to `Single`

bergman (approved):
This seems to change very little in the API (only adds a few things), breaks nothing and gives a nicer experience for the single-value input/output 👍

rafaelraposospot (changes requested):

flytekit-java/src/main/java/org/flyte/flytekit/MonoVariableSdkType.java
narape: Right, I will follow Jocke advice in the name

bergman:
> @andresgomezfrr I guess your question comes from passing the name of the property as string, but take a look at the SumTask and now with the latest commit to the FibbonacciWorkflow. For example
> 
> ```scala
>     val fib2: SdkBindingData[Long] = builder
>       .apply("fib-2", new SumTask(), SumTaskInput(input.fib0, input.fib1))
>       .getOutputs
> ```
> 
> notice now `getOutputs` return the `SdkBindingData[Long]` directly, no need to call any `.c()` method. This is why the name of the variable could be defined as string, because it would be hidden from the wiring code, yet still present internally in the promises of the idl

I think this is a little confusing, the method is named `getOutputs` but it returns a single value. This is why I wasn't so keen on naming that method that in the first place 😂

pablocasares (approved):

rafaelraposospot (approved):

PR #209, opened 2023-03-10
Add ability to fetch lp with version

sonjaer:
# TL;DR
Factory method to create remote launch plan with fixed version

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added - NA there are already tests for it
 - [ ] Code documentation added NA
 - [ ] Any pending items have an associated Issue NA

### DIFF
--- a/flytekit-java/src/main/java/org/flyte/flytekit/SdkRemoteLaunchPlan.java
+++ b/flytekit-java/src/main/java/org/flyte/flytekit/SdkRemoteLaunchPlan.java
@@ -91,6 +91,34 @@ public abstract class SdkRemoteLaunchPlan<InputT, OutputT> extends SdkTransform<
         .build();
   }
 
+  /**
+   * Create a remote launch plan, a reference to a launch plan that have been deployed previously.
+   *
+   * @param domain the domain of the remote launch plan
+   * @param project the project of the remote launch plan
+   * @param name the name of the remote launch plan
+   * @param version of the remote launch plan
+   * @param inputs the {@link SdkType} for the inputs of the remote launch plan
+   * @param outputs the {@link SdkType} for the outputs of the remote launch plan
+   * @return the remote launch plan
+   */
+  public static <InputT, OutputT> SdkRemoteLaunchPlan<InputT, OutputT> create(
+          String domain,
+          String project,
+          String name,
+          String version,
+          SdkType<InputT> inputs,
+          SdkType<OutputT> outputs) {
+    return SdkRemoteLaunchPlan.<InputT, OutputT>builder()
+            .domain(domain)
+            .project(project)
+            .name(name)
+            .version(version)
+            .inputs(inputs)
+            .outputs(outputs)
+            .build();
+  }
+
   /** {@inheritDoc} */
   @Override
   public String getName() {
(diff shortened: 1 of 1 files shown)


honnix:
Why is this needed given user can use the Builder directly?

sonjaer:
> Why is this needed given user can use the Builder directly?

It is more user friendly

honnix:
If we really want to offer the factor method, let's at least refactor the existing one:

```java
  public static <InputT, OutputT> SdkRemoteLaunchPlan<InputT, OutputT> create(
      String domain,
      String project,
      String name,
      SdkType<InputT> inputs,
      SdkType<OutputT> outputs) {
    return create(domain, project, name, null, inputs, outputs);
  }
```

flytekit-java/src/main/java/org/flyte/flytekit/SdkRemoteLaunchPlan.java
```diff
@@ -82,10 +83,32 @@ public static <InputT, OutputT> SdkRemoteLaunchPlan<InputT, OutputT> create(
...
+   * @param version of the remote launch plan
+   * @param inputs the {@link SdkType} for the inputs of the remote launch plan
+   * @param outputs the {@link SdkType} for the outputs of the remote launch plan
+   * @return the remote launch plan
+   */
+  public static <InputT, OutputT> SdkRemoteLaunchPlan<InputT, OutputT> create(
+      String domain,
+      String project,
+      String name,
+      String version,
```
honnix, line 104:
```suggestion
      @Nullable String version,
```

flytekit-java/src/main/java/org/flyte/flytekit/SdkRemoteLaunchPlan.java
```diff
@@ -82,10 +83,32 @@ public static <InputT, OutputT> SdkRemoteLaunchPlan<InputT, OutputT> create(
...
+    return create(domain, project, name, null, inputs, outputs);
+  }
+
+  /**
+   * Create a remote launch plan, a reference to a launch plan that have been deployed previously.
+   *
+   * @param domain the domain of the remote launch plan
+   * @param project the project of the remote launch plan
+   * @param name the name of the remote launch plan
+   * @param version of the remote launch plan
```
honnix, line 95:
```suggestion
   * @param version of the remote launch plan, defaults to the latest version if null
```

honnix:
> > Why is this needed given user can use the Builder directly?
> 
> It is more user friendly

I understand that rationale, and we should also be careful when adding factory method because it is very hard to deprecate.

sonjaer:
> > > Why is this needed given user can use the Builder directly?
> > 
> > 
> > It is more user friendly
> 
> I understand that rationale, and we should also be careful when adding factory method because it is very hard to deprecate.

ok, so its better to expose the builder directly? is that easier to manage/deprecate?

honnix (approved):

honnix:
> > > > Why is this needed given user can use the Builder directly?
> > > 
> > > 
> > > It is more user friendly
> > 
> > 
> > I understand that rationale, and we should also be careful when adding factory method because it is very hard to deprecate.
> 
> ok, so its better to expose the builder directly? is that easier to manage/deprecate?

That is hard to say. My point is more about minimizing the surface we expose to users, and avoiding offering competing ways to achieve the same. This addition is too small to justify that reasoning, so I'm fine we ship it.

honnix:
Now I see the build is not public. :( We have many things with weird visibility in the code base and we should probably revisit those.

honnix:
Should we apply the same treatment to https://github.com/flyteorg/flytekit-java/blob/master/flytekit-java/src/main/java/org/flyte/flytekit/SdkRemoteTask.java?

PR #212, opened 2023-03-17
Fix dynamic workflows serialization

narape:
# TL;DR
Fix dynamic workflow task serialization. We broke dynamic workflows when we added strong types to workflows. 

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added

## Complete description
Dynamic workflows output a [DynamicJobSpec](https://github.com/flyteorg/flyteidl/blob/master/protos/flyteidl/core/dynamic_job.proto#L12) that gets written in the propeller bucket as `futures.pb`, then propeller read this file and executes the workflow.

The problem was that we changes how we register outputs with the strong types and this spec proto didn't contain the binding for the outputs, then the written file was missing this information and propeller failed and retried the task until the retry counts reach its limit,

We didn't notice the breakage because we didn't contain unit tests for that (fixed) and the integration test was disabled (now enabled as well)

## Tracking Issue
NA

## Follow-up issue
_NA_

### DIFF
--- a/flytekit-java/src/main/java/org/flyte/flytekit/SdkDynamicWorkflowTaskRegistrar.java
+++ b/flytekit-java/src/main/java/org/flyte/flytekit/SdkDynamicWorkflowTaskRegistrar.java
@@ -20,6 +20,7 @@ import static java.util.Collections.emptyMap;
 import static java.util.stream.Collectors.toUnmodifiableList;
 
 import com.google.auto.service.AutoService;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


flytekit-java/src/main/java/org/flyte/flytekit/SdkDynamicWorkflowTaskRegistrar.java
```diff
@@ -82,12 +83,21 @@ public DynamicJobSpec run(Map<String, Literal> inputs) {
       SdkWorkflowBuilder builder = new SdkWorkflowBuilder();
 
       InputT value = sdkDynamicWorkflow.getInputType().fromLiteralMap(inputs);
-      sdkDynamicWorkflow.run(builder, value);
+      var output = sdkDynamicWorkflow.run(builder, value);
```
honnix, line 86: Have we agreed to start using `var` for everything? If so, maybe should do a big refactoring PR later.

andresgomezfrr (approved):

pablocasares (approved):

flytekit-java/src/main/java/org/flyte/flytekit/SdkDynamicWorkflowTaskRegistrar.java
bergman: I don't think so, let's revert/explicit type

PR #216, opened 2023-03-23
More flexible

andresgomezfrr:
# TL;DR
Add support to create nodes using a transform `SdkTransform[InputT, OutputT]` and an input `Map[String, SdkBindingData[_]]`.

This was supported previously and we removed this feature with the strongly-types interface.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Tracking Issue


flytekit-java/src/test/java/org/flyte/flytekit/SdkWorkflowBuilderTest.java
narape, line 1: Add a test where you pass a map with more inputs than needed and a second one where you pass a map with fewer inputs than needed by the tasks

flytekit-scala_2.13/src/main/scala/org/flyte/flytekitscala/SdkScalaWorkflow.scala
narape, line 1: I just realized that we don't have tests for this class

narape (commented):
A good way to recover some dynamic typing

flytekit-scala_2.13/src/main/scala/org/flyte/flytekitscala/SdkScalaWorkflow.scala
narape: There is very little code in this class. We can disregard my comment

narape (approved):

PR #217, opened 2023-03-24
No spying behaviour for testing

honnix:
## _Read then delete this section_

_- Make sure to use a concise title for the pull-request._

_- Use #patch, #minor or #major in the pull-request title to bump the corresponding version. Otherwise, the patch version
will be bumped. [More details](https://github.com/marketplace/actions/github-tag-bump)_

# TL;DR
_Please replace this text with a description of what this PR accomplishes._

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added
 - [ ] Code documentation added
 - [ ] Any pending items have an associated Issue

## Complete description
 _How did you fix the bug, make the feature etc. Link to any design docs etc_

## Tracking Issue
_Remove the '*fixes*' keyword if there will be multiple PRs to fix the linked issue_

fixes https://github.com/flyteorg/flyte/issues/<number>

## Follow-up issue
_NA_
OR
_https://github.com/flyteorg/flyte/issues/<number>_

PR #214, opened 2023-03-23
Allow to configure to enable/disable runnable task during testing

andresgomezfrr:
## _Read then delete this section_

_- Make sure to use a concise title for the pull-request._

_- Use #patch, #minor or #major in the pull-request title to bump the corresponding version. Otherwise, the patch version
will be bumped. [More details](https://github.com/marketplace/actions/github-tag-bump)_

# TL;DR
_Please replace this text with a description of what this PR accomplishes._

This PR does the following:
* Disable runnable task execution in the SdkTestingExecutor by default.
* Now you can add explicitly if you want to allow to the SdkTestingExecutor runs runnable task locally.
   * Could be configurable by all inputs of a specific task.
   ```java
     .withEnableRunnableFor(new SumTask())
   ```
   * Could be configurable by only a group of inputs of a specific task.
   ```java
   .withEnableRunnableForInput(new SumTask(), )
   ```

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added
 - [ ] Code documentation added
 - [ ] Any pending items have an associated Issue

## Complete description
 _How did you fix the bug, make the feature etc. Link to any design docs etc_

## Tracking Issue
_Remove the '*fixes*' keyword if there will be multiple PRs to fix the linked issue_

fixes https://github.com/flyteorg/flyte/issues/<number>

## Follow-up issue
_NA_
OR
_https://github.com/flyteorg/flyte/issues/<number>_

PR #218, opened 2023-03-25
Explicitly fall back to real test execution when testing

honnix:
# TL;DR

More or less make #217 complete.

Closes #214 
Closes #217 

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
* If a task is not mocked via input matching
  * If lambda is not null, we run the lambda (lambda being either the real thing or provided)
  * Otherwise we throw
* If a task is mocked
  * If inputs match, we return output
  * If it is mocked with a provided lambda, we run the lambda
  * Otherwise, we throw

## Tracking Issue
https://github.com/flyteorg/flyte/issues/3522

## Follow-up issue
_NA_

### DIFF
--- a/flytekit-testing/src/main/java/org/flyte/flytekit/testing/TestingRunnableLaunchPlan.java
+++ b/flytekit-testing/src/main/java/org/flyte/flytekit/testing/TestingRunnableLaunchPlan.java
@@ -35,12 +35,14 @@ public class TestingRunnableLaunchPlan<InputT, OutputT>
       SdkType<InputT> inputType,
       SdkType<OutputT> outputType,
       Function<InputT, OutputT> runFn,
+      boolean runFnProvided,
       Map<InputT, OutputT> fixedOutputs) {
     super(
         launchPlanId,
         inputType,
         outputType,
         runFn,
+        runFnProvided,
         fixedOutputs,
         TestingRunnableLaunchPlan::new,
         "launch plan",
@@ -52,6 +54,7 @@ public class TestingRunnableLaunchPlan<InputT, OutputT>
     PartialLaunchPlanIdentifier launchPlanId =
         PartialLaunchPlanIdentifier.builder().name(name).build();
 
-    return new TestingRunnableLaunchPlan<>(launchPlanId, inputType, outputType, null, emptyMap());
+    return new TestingRunnableLaunchPlan<>(
+        launchPlanId, inputType, outputType, null, false, emptyMap());
   }
 }
--- a/flytekit-testing/src/test/java/org/flyte/flytekit/testing/FibonacciWorkflowTest.java
+++ b/flytekit-testing/src/test/java/org/flyte/flytekit/testing/FibonacciWorkflowTest.java
@@ -73,6 +73,7 @@ public class FibonacciWorkflowTest {
                 new SumTask(),
                 SumInput.create(SdkBindingDataFactory.of(3L), SdkBindingDataFactory.of(5L)),
                 SumOutput.create(SdkBindingDataFactory.of(42L)))
+            .withTask(new SumTask(), new SumTask()::run)
             .execute();
 
     assertThat(result.getIntegerOutput("fib2"), equalTo(2L));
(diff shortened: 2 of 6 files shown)


flytekit-testing/src/main/java/org/flyte/flytekit/testing/TestingRunnableLaunchPlan.java
```diff
@@ -35,12 +35,14 @@
...
       SdkType<OutputT> outputType,
       Function<InputT, OutputT> runFn,
+      boolean runFnProvided,
       Map<InputT, OutputT> fixedOutputs) {
     super(
         launchPlanId,
         inputType,
         outputType,
         runFn,
+        runFnProvided,
```
honnix, line 45: Introducing this because otherwise it not possible to tell whether user has provided a `runFn`, because `runFn` could either point to the real task `run` or a user provided one.

flytekit-testing/src/test/java/org/flyte/flytekit/testing/FibonacciWorkflowTest.java
```diff
@@ -73,6 +73,7 @@ public void testWithTaskOutput_runnableTask() {
                 new SumTask(),
                 SumInput.create(SdkBindingDataFactory.of(3L), SdkBindingDataFactory.of(5L)),
                 SumOutput.create(SdkBindingDataFactory.of(42L)))
+            .withTask(new SumTask(), new SumTask()::run)
```
honnix, line 76: This is how users can explicitly enable the fallback behaviour.

bergman:
> > It would be nice to have some more tests around e.g. how mocks are reused, what happens if they are unused and such things like how mockito and others behave, but we can do that in another PR.
> 
> I think they are more or less covered by https://github.com/flyteorg/flytekit-java/blob/bd61fe3e2b9e350b5b7330832a403f064cf3f6ef/flytekit-testing/src/test/java/org/flyte/flytekit/testing/TestingRunnableNodeTest.java. The coverage also looks good (TestingRunnableNode.java is fully covered). What else do you think we should add?

It seems these properties of mocks are unexplored and they all relate to the number of times they are used. They all do the "wrong" or at least unexpected thing compared to e.g. mockito, so I think we should at some point change their behaviour:
- unused mocks (should throw but does nothing)
- duplicate mocks (should maybe throw but does nothing, maybe caught as unused mock above)
- same mock used multiple times (not sure if this is by default in mockito or if it's part of `verify` but should at least be configurable)

honnix:
I think Mockito works in this way:

* unused mocks (should throw but does nothing)

Mockito defaults to fail the test, but it can be disabled by marking a mock as `enient`.

* duplicate mocks (should maybe throw but does nothing, maybe caught as unused mock above)

This is more or less the same as "unused mocks".

* same mock used multiple times (not sure if this is by default in mockito or if it's part of verify but should at least be configurable)

This is an expected usage, and yes it is reflected in the `verify` call to use different `times()`.

honnix:
Hmm, looks like we have some flaky integration tests.

bergman:
> I think Mockito works in this way:
> 
> * unused mocks (should throw but does nothing)
> 
> Mockito defaults to fail the test, but it can be disabled by marking a mock as `enient`.
> 
> * duplicate mocks (should maybe throw but does nothing, maybe caught as unused mock above)
> 
> This is more or less the same as "unused mocks".
> 
> * same mock used multiple times (not sure if this is by default in mockito or if it's part of verify but should at least be configurable)
> 
> This is an expected usage, and yes it is reflected in the `verify` call to use different `times()`.

Agreed on all accounts except that duplicate mocks should be explicitly forbidden since order of execution is not guaranteed in a workflow (or maybe it is, I just kinda assumed there wasn't because it's usually an implementation detail in what order a graph is traversed). Implemented tests and validations for these in #222

PR #238, opened 2023-07-31
Add custom JAVA_TOOL_OPTIONS by SdkRunnableTask

andresgomezfrr:
## _Read then delete this section_

Allow configuring custom JAVA_TOOL_OPTIONS by SdkRunnableTask

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added


flytekit-api/src/main/java/org/flyte/api/v1/RunnableTask.java
```diff
@@ -37,4 +39,8 @@ default Resources getResources() {
     // FIXME default only for backwards-compatibility, remove in 0.3.x
     return Resources.builder().build();
   }
+
+  default List<String> getCustomJavaToolOptions() {
```
honnix, line 43: I don't think we need default implementation here. Actually all the other default methods in this class should have been removed long time ago. We do not need to touch other in this PR, but let's not add another "FIXME".

jflyte/src/main/java/org/flyte/jflyte/ProjectClosure.java
```diff
@@ -496,13 +496,25 @@ private static TaskTemplate createTaskTemplate(Task task, Container container) {
...
+
+    Resources resources = task.getResources();
     Map<ResourceName, String> limits = resources.limits();
-    if (limits == null || !limits.containsKey(ResourceName.MEMORY)) {
+    if (limits != null && limits.containsKey(ResourceName.MEMORY)) {
+      String maxMemory = asJavaQuantity(limits.get(ResourceName.MEMORY));
+      javaToolOptions.add("-Xmx" + maxMemory);
+    }
+
+    if (!task.getCustomJavaToolOptions().isEmpty()) {
```
honnix, line 509: There is no need to do a check here.

honnix (commented):
Minor comments, otherwise LGTM.

honnix:
Please create an issue and reference to it for visibility, as this is a pretty useful new feature.

flytekit-api/src/main/java/org/flyte/api/v1/RunnableTask.java
```diff
@@ -39,8 +39,4 @@ default Resources getResources() {
     // FIXME default only for backwards-compatibility, remove in 0.3.x
     return Resources.builder().build();
   }
-
-  default List<String> getCustomJavaToolOptions() {
```
honnix: This method is still needed. It's just we don't need to give a default implementation.

andresgomezfrr: yeah! You're right. I don't know what I was thinking... 🙈

flytekit-java/src/main/java/org/flyte/flytekit/SdkRunnableTaskRegistrar.java
```diff
@@ -110,6 +111,11 @@ public String getName() {
     public Resources getResources() {
       return sdkTask.getResources().toIdl();
     }
+
+    @Override
+    public List<String> getCustomJavaToolOptions() {
+      return sdkTask.getCustomJavaToolOptions();
```
honnix, line 117: Hmm, this is actually problematic. `flytekit-java` is shipped from user code space, which means for existing code, `getCustomJavaToolOptions` does not exist in `SdkRunnableTask`. So here for now we need to use reflection to check existence of this method, until we migrate all users.

honnix (changes requested):
This breaks backward compatibility. Please check my comment https://github.com/flyteorg/flytekit-java/pull/238/files#r1279358563.

flytekit-java/src/main/java/org/flyte/flytekit/SdkRunnableTaskRegistrar.java
andresgomezfrr: You're right... this is the case where a user uses an old version of flytekit-java and tries to register with a new version of it, right?

honnix: Let's me think about it. It may not be a problem. I need to check how jflyte works.

honnix: Forgot about my previous comment. I somehow thought this code was in jflyte module.

I think we will need to have default implementation in `RunnableTask` after all. `jflyte` will use `flytekit-java` module from user code space when registering because of ChildFirstClassLoader, so when processing old version where `RunnableTaskImpl` does not have `getCustomJavaToolOptions`, `ProjectClosure` will fail when calling `task.getCustomJavaToolOptions()`, or maybe even earlier because old `RunnableTaskImpl` does not really implement new `RunnableTask` if there is no default implementation.

It might come as a common rule that there must be default implementation in `RunnableTask` whenever we extend it, so the existing `FIXME`s are not telling the truth.

andresgomezfrr: so.. in this case, should I remove all the FIXME comments?

andresgomezfrr: I tested it using the old flytekit-java in the user repo and registered the task with the new flytekit-java container and with the default property, everything worked fine.

flytekit-api/src/main/java/org/flyte/api/v1/RunnableTask.java
honnix: This comment is not true. Please refer to https://github.com/flyteorg/flytekit-java/pull/238/files#r1279405058

flytekit-java/src/main/java/org/flyte/flytekit/SdkRunnableTaskRegistrar.java
honnix: > so.. in this case, should I remove all the FIXME comments?

I think it makes sense. It is rather simple to keep the defaults than asking users to migrate, and the migration itself doesn't bring much value anyway.

honnix: > I tested it using the old flytekit-java in the user repo and registered the task with the new flytekit-java container and with the default property, everything worked fine.

Great. Sorry I gave wrong direction in an early comment. I did not think carefully about those "FIXME"s .

andresgomezfrr: Don't worry! so.. looks good to you the PR now?

honnix (approved):

honnix:
There are some test failures.

andresgomezfrr:
> There are some test failures.

yes... I just saw this.. checking

andresgomezfrr:
> There are some test failures.

I think that this test return a duplicated 😮 

```
[ERROR] Failures: 
[ERROR]   SdkNodeNamePolicyTest.nextNodeIdShouldReturnUniquePrefixForDistinctPolicies:71 returned duplicated prefix
Expected: a collection with size <9>
     but: collection size was <10>
```


```java
  @Test
  void nextNodeIdShouldReturnUniquePrefixForDistinctPolicies() {
    var prefixes =
        Stream.generate(SdkNodeNamePolicy::new)
            .limit(10)
            .map(SdkNodeNamePolicy::nextNodeId)
            .map(SdkNodeNamePolicyTest::prefix)
            .collect(toList());
    var uniquePrefixes = Set.copyOf(prefixes);

    assertThat("returned duplicated prefix", prefixes, hasSize(uniquePrefixes.size()));
  }
  ```

PR #244, opened 2023-09-25
Extract public

honnix:
# TL;DR
Extract jflyte utility classes into a separated module.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description

 This is to make it easier to extend jflyte or replacing certain subcommands, otherwise things have to be done in `org.flyte.jflyte` packages.

Note that not everything in jflyte-utils is public because we only want to open up things we know for sure are already being used and we can open up others when needed.

## Tracking Issue
_NA_

## Follow-up issue
_NA_


jflyte-utils/pom.xml
```diff
@@ -0,0 +1,182 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Copyright 2021 Flyte Authors.
```
bergman, line 3:
```suggestion
  Copyright 2023 Flyte Authors.
```
Unless there is some specific requirement that it this license section is the exact same everywhere?

bergman (approved):
Had only a nitpick.

jflyte-utils/pom.xml
bergman: Ah, I was afraid it was something like this. Thanks for confirming!

PR #245, opened 2023-09-26
Fix license header

honnix:
# TL;DR
Fix license header by using a year parameter

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
With this, we will not be forced to use 2021 forever.

## Tracking Issue

_NA_

## Follow-up issue
_NA_

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -418,7 +418,7 @@
               <licenseHeader>
                 <!-- Specify either content or file, but not both -->
                 <content><![CDATA[/*
- * Copyright 2021 Flyte Authors
+ * Copyright $YEAR Flyte Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -440,40 +440,13 @@
               </googleJavaFormat>
               <removeUnusedImports />
             </java>
-            <xml>
```
honnix, line 443: I don't know how we came up with this `<xml>` configuration but it doesn't seems to be supported by spotless at all.

honnix:
We could also retrospectively updating all headers according to git commit history but that would be a massive change. Do we want to do this?

It is also worth noting that spotless license header thing does not support `package-info.java`, so those files we can only do manually.

pom.xml
```diff
@@ -418,7 +418,7 @@
               <licenseHeader>
                 <!-- Specify either content or file, but not both -->
                 <content><![CDATA[/*
- * Copyright 2021 Flyte Authors
+ * Copyright $YEAR Flyte Authors
```
honnix, line 421: Why does this have no `.` :( To fix that we will need to update all Java files.

andresgomezfrr (approved):

PR #248, opened 2023-09-27
Support remote in dynamic

honnix:
# TL;DR
This is an attempt to support remote tasks/launch plans in dynamic workflows.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
I don't see a good reason why we don't support the remote in dynamic setup, while we already have code doing the latest lookup etc.

## Tracking Issue
_NA_

## Follow-up issue
_NA_

### DIFF
--- a/flytekit-examples/src/main/java/org/flyte/examples/DynamicFibonacciWorkflowTask.java
+++ b/flytekit-examples/src/main/java/org/flyte/examples/DynamicFibonacciWorkflowTask.java
@@ -16,6 +16,9 @@
  */
 package org.flyte.examples;
 
+import static org.flyte.examples.FlyteEnvironment.DOMAIN;
+import static org.flyte.examples.FlyteEnvironment.PROJECT;
+
 import com.google.auto.service.AutoService;
 import com.google.auto.value.AutoValue;
 import com.google.errorprone.annotations.Var;
@@ -23,6 +26,9 @@ import org.flyte.examples.SumTask.SumInput;
 import org.flyte.flytekit.SdkBindingData;
 import org.flyte.flytekit.SdkBindingDataFactory;
 import org.flyte.flytekit.SdkDynamicWorkflowTask;
+import org.flyte.flytekit.SdkNode;
+import org.flyte.flytekit.SdkRemoteTask;
+import org.flyte.flytekit.SdkTypes;
 import org.flyte.flytekit.SdkWorkflowBuilder;
 import org.flyte.flytekit.jackson.JacksonSdkType;
 
@@ -59,11 +65,23 @@ public class DynamicFibonacciWorkflowTask
     } else if (input.n().get() == 0) {
       return Output.create(SdkBindingDataFactory.of(0));
     } else {
+      SdkNode<Void> hello =
+          builder.apply(
+              "hello",
+              SdkRemoteTask.create(
+                  DOMAIN,
+                  PROJECT,
+                  HelloWorldTask.class.getName(),
+                  SdkTypes.nulls(),
+                  SdkTypes.nulls()));
       @Var SdkBindingData<Long> prev = SdkBindingDataFactory.of(0);
       @Var SdkBindingData<Long> value = SdkBindingDataFactory.of(1);
       for (int i = 2; i <= input.n().get(); i++) {
         SdkBindingData<Long> next =
-            builder.apply("fib-" + i, new SumTask(), SumInput.create(value, prev)).getOutputs();
+            builder
+                .apply(
+                    "fib-" + i, new SumTask().withUpstreamNode(hello), SumInput.create(value, prev))
+                .getOutputs();
         prev = value;
         value = next;
       }
--- a/flytekit-testing/src/main/java/org/flyte/flytekit/testing/SdkTestingExecutor.java
+++ b/flytekit-testing/src/main/java/org/flyte/flytekit/testing/SdkTestingExecutor.java
@@ -42,9 +42,11 @@ import org.flyte.api.v1.Variable;
 import org.flyte.api.v1.WorkflowNode;
 import org.flyte.api.v1.WorkflowNode.Reference;
 import org.flyte.api.v1.WorkflowTemplate;
+import org.flyte.flytekit.SdkDynamicWorkflowTask;
 import org.flyte.flytekit.SdkRemoteLaunchPlan;
 import org.flyte.flytekit.SdkRemoteTask;
 import org.flyte.flytekit.SdkRunnableTask;
+import org.flyte.flytekit.SdkTransform;
 import org.flyte.flytekit.SdkType;
 import org.flyte.flytekit.SdkWorkflow;
 import org.flyte.localengine.ExecutionContext;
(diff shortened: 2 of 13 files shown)


jflyte/src/main/java/org/flyte/jflyte/ExecuteDynamicWorkflow.java
```diff
@@ -235,69 +239,6 @@ static DynamicJobSpec rewrite(
...
-  static class DynamicWorkflowIdentifierRewrite extends WorkflowNodeVisitor {
-    private final ExecutionConfig config;
-
-    DynamicWorkflowIdentifierRewrite(ExecutionConfig config) {
-      this.config = config;
-    }
-
-    @Override
-    protected PartialTaskIdentifier visitTaskIdentifier(PartialTaskIdentifier value) {
-      if (value.project() == null && value.domain() == null && value.version() == null) {
```
honnix, line 247: So the current logic is, we only support task without any pdv coordinate, which means tasks in the workflow definition itself. For a remote entity, we do have pd(v).

honnix: Take an example from `IdentifierRewrite`.

```java
  PartialTaskIdentifier apply(PartialTaskIdentifier taskId) {
    String name = Preconditions.checkNotNull(taskId.name(), "name is null");
    String project = coalesce(taskId.project(), project());
    String domain = coalesce(taskId.domain(), domain());
    String version =
        coalesce(
            taskId.version(),
            () ->
                taskId.project() == null
                    ? version()
                    : getLatestTaskVersion(
                        /* project= */ project, /* domain= */ domain, /* name= */ name));
    return PartialTaskIdentifier.builder()
        .name(name)
        .domain(domain)
        .project(project)
        .version(version)
        .build();
  }
```

This means:

* we take task name from task id, which is the same as `DynamicWorkflowIdentifierRewrite`
* we take project if task id has it and in this case it is a remote one; but if it is a task in the workflow itself, this should be `null`, and this should cover the existing path in `DynamicWorkflowIdentifierRewrite`
* the same applies to domain
* we take version if task id has it and in this case it is a remote one with resolved (or hardcoded) version; but if it is a task in the workflow itself, this should be `null` and the same for project, so this should cover the existing path in `DynamicWorkflowIdentifierRewrite`; but if project is not null we will try to do a latest lookup to resolve the remote one

With this, we should both support local and remote tasks. The same applies to remote launch plan, and subworkflow.

honnix: There might be some corner cases that we may eventually need to write a new visitor, but I honestly don't see any blocker.

integration-tests/src/test/java/org/flyte/utils/JFlyteContainer.java
```diff
@@ -30,8 +30,6 @@ class JFlyteContainer extends GenericContainer<JFlyteContainer> {
   static final String IMAGE_NAME;
   static final Map<String, String> envVars =
       ImmutableMap.<String, String>builder()
-          .put("FLYTE_PLATFORM_URL", "flyte:30081")
```
honnix, line 33: These are already when building jflyte image.

flytekit-examples/src/main/java/org/flyte/examples/DynamicFibonacciWorkflowTask.java
```diff
@@ -59,11 +65,23 @@ public Output run(SdkWorkflowBuilder builder, Input input) {
     } else if (input.n().get() == 0) {
       return Output.create(SdkBindingDataFactory.of(0));
     } else {
+      SdkNode<Void> hello =
+          builder.apply(
+              "hello",
+              SdkRemoteTask.create(
```
honnix, line 71: This is the remote task we reference from a dynamic workflow task.

flytekit-examples/src/main/java/org/flyte/examples/DynamicFibonacciWorkflowTask.java
```diff
@@ -59,11 +65,23 @@ public Output run(SdkWorkflowBuilder builder, Input input) {
...
+                  SdkTypes.nulls(),
+                  SdkTypes.nulls()));
       @Var SdkBindingData<Long> prev = SdkBindingDataFactory.of(0);
       @Var SdkBindingData<Long> value = SdkBindingDataFactory.of(1);
       for (int i = 2; i <= input.n().get(); i++) {
         SdkBindingData<Long> next =
-            builder.apply("fib-" + i, new SumTask(), SumInput.create(value, prev)).getOutputs();
+            builder
+                .apply(
+                    "fib-" + i, new SumTask().withUpstreamNode(hello), SumInput.create(value, prev))
```
honnix, line 83: Just to make a connection for fun, as it is not required.

jflyte/src/main/java/org/flyte/jflyte/ExecuteDynamicWorkflow.java
```diff
@@ -198,15 +197,24 @@ private void execute() {
...
   static DynamicJobSpec rewrite(
-      ExecutionConfig config,
+      Config config,
+      ExecutionConfig executionConfig,
       DynamicJobSpec spec,
       Map<TaskIdentifier, TaskTemplate> taskTemplates,
       Map<WorkflowIdentifier, WorkflowTemplate> workflowTemplates) {
 
-    DynamicWorkflowIdentifierRewrite rewrite = new DynamicWorkflowIdentifierRewrite(config);
+    WorkflowNodeVisitor workflowNodeVisitor =
```
honnix, line 206: This is the core change.

integration-tests/src/test/java/org/flyte/utils/FlyteSandboxContainer.java
```diff
@@ -33,7 +33,7 @@
 
 public class FlyteSandboxContainer extends GenericContainer<FlyteSandboxContainer> {
 
-  public static final String IMAGE_NAME = "ghcr.io/flyteorg/flyte-sandbox:v1.1.0";
+  public static final String IMAGE_NAME = "ghcr.io/flyteorg/flyte-sandbox:v1.9.1";
```
honnix, line 36: Not required by this PR, but it's nice to be on latest version.

jflyte/pom.xml
```diff
@@ -200,7 +200,7 @@
             <environment>
               <FLYTE_INTERNAL_MODULE_DIR>/jflyte/modules</FLYTE_INTERNAL_MODULE_DIR>
               <FLYTE_INTERNAL_IMAGE>${docker.image}:${docker.tag}</FLYTE_INTERNAL_IMAGE>
-              <FLYTE_PLATFORM_URL>flyte:30081</FLYTE_PLATFORM_URL>
+              <FLYTE_PLATFORM_URL>flyteadmin.flyte.svc.cluster.local:81</FLYTE_PLATFORM_URL>
```
honnix, line 203: This is required for remote lookup to work because the task container running in k3 will need to talk to flyteadmin and using service DNS name is the best option.

Strictly speaking this is a breaking change but it is extremely unlikely that the existing config `flyte:30081` would work for any users anyway and they have likely already customized it.

flytekit-testing/src/main/java/org/flyte/flytekit/testing/SdkTestingExecutor.java
```diff
@@ -321,20 +323,32 @@ public <T> SdkTestingExecutor withFixedInputs(SdkType<T> type, T value) {
...
       SdkRunnableTask<InputT, OutputT> task, InputT input, OutputT output) {
+    return withTaskOutput0(task, input, output);
+  }
+
+  public <InputT, OutputT> SdkTestingExecutor withTaskOutput(
+      SdkDynamicWorkflowTask<InputT, OutputT> task, InputT input, OutputT output) {
+    return withTaskOutput0(task, input, output);
+  }
+
+  public <InputT, OutputT> SdkTestingExecutor withTaskOutput(
```
honnix, line 334: The diff is weird, but there is no change made to `withTaskOutput(SdkRemoteTask, ...)`,

andresgomezfrr (approved):

PR #254, opened 2023-09-29
Fetch task template in dynamic workflow task

honnix:
# TL;DR
Fetch task template when encountering remote task in dynamic workflow.

## Type
 - [x] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
#248 is not sufficient because in order to build dynamic job, we need task template, so in this PR we fetch task template for remote workflow.

## Tracking Issue
_NA_

## Follow-up issue
_NA_

### DIFF
--- a/jflyte-utils/src/main/java/org/flyte/jflyte/utils/FlyteAdminClient.java
+++ b/jflyte-utils/src/main/java/org/flyte/jflyte/utils/FlyteAdminClient.java
@@ -25,7 +25,6 @@ import flyteidl.admin.ExecutionOuterClass;
 import flyteidl.admin.LaunchPlanOuterClass;
 import flyteidl.admin.TaskOuterClass;
 import flyteidl.admin.WorkflowOuterClass;
-import flyteidl.core.IdentifierOuterClass;
 import flyteidl.service.AdminServiceGrpc;
 import io.grpc.Channel;
 import io.grpc.ClientInterceptor;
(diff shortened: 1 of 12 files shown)


jflyte-utils/src/main/java/org/flyte/jflyte/utils/FlyteAdminClient.java
```diff
@@ -185,34 +184,38 @@ public TaskIdentifier fetchLatestTaskId(NamedEntityIdentifier taskId) {
     return fetchLatestResource(
         taskId,
         request -> stub.listTasks(request).getTasksList(),
-        TaskOuterClass.Task::getId,
```
honnix, line 188: This double-function doesn't seem to be necessary.

jflyte-utils/src/main/java/org/flyte/jflyte/utils/ProjectClosure.java
```diff
@@ -374,8 +375,10 @@ public static Map<WorkflowIdentifier, WorkflowTemplate> collectSubWorkflows(
         .collect(toUnmodifiableMap());
   }
 
-  public static Map<TaskIdentifier, TaskTemplate> collectTasks(
-      List<Node> rewrittenNodes, Map<TaskIdentifier, TaskTemplate> allTasks) {
+  public static Map<TaskIdentifier, TaskTemplate> collectDynamicWorkflowTasks(
```
honnix, line 378: Rename it to be more specific. This is not used by anything else.

andresgomezfrr (approved):

PR #257, opened 2023-10-06
Describe known inputs when input type not found in TestingRunnableNode

acet:
# TL;DR
When attempting a workflow test, there can be failures setting up the test when incorrect parameters are used. It is not immediately obvious from the current exception message where the difference is between expected and actual.
Here I have included the known inputs that are checked when setting up the test so that the user has a better idea of what is wrong. 

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
I had to create workflow tests with several LaunchPlanOutput/TaskOutput for bigquery, gcs, dataflow jobs, etc. Because of the large number of parameters I was using it was helpful for me to debug `TestingRunnableNode` when I got the initial error message to compare my input with the known inputs. I thought a more detailed error message could be of use to someone facing the same problem. 

## Tracking Issue

## Follow-up issue
_NA_
OR
_https://github.com/flyteorg/flyte/issues/<number>_

### DIFF
--- a/flytekit-testing/src/main/java/org/flyte/flytekit/testing/TestingRunnableNode.java
+++ b/flytekit-testing/src/main/java/org/flyte/flytekit/testing/TestingRunnableNode.java
@@ -20,7 +20,9 @@ import static java.util.Objects.requireNonNull;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
 import java.util.function.Function;
+import java.util.stream.Collectors;
 import org.flyte.api.v1.Literal;
 import org.flyte.api.v1.PartialIdentifier;
 import org.flyte.api.v1.RunnableNode;
@@ -84,7 +86,7 @@ public abstract class TestingRunnableNode<
   public Map<String, Literal> run(Map<String, Literal> inputs) {
     InputT input = inputType.fromLiteralMap(inputs);
 
-    if (fixedOutputs.size() == 0) {
+    if (fixedOutputs.isEmpty()) {
       // No mocking via input matching, either run the real thing or run the provided lambda
       if (runFn != null) {
         return outputType.toLiteralMap(runFn.apply(input));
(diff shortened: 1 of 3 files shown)


flytekit-testing/src/main/java/org/flyte/flytekit/testing/TestingRunnableNode.java
```diff
@@ -101,14 +103,31 @@ public Map<String, Literal> run(Map<String, Literal> inputs) {
...
+                System.lineSeparator(),
+                knownInputsString(),
+                System.lineSeparator(),
+                testingSuggestion);
 
     // Not matching inputs and there is nothing to run
     throw new IllegalArgumentException(message);
   }
 
+  private String knownInputsString() {
```
AndersonReyes, line 120: Just being nosy because I want this improvement as well but I wonder if we should filter the fixedOutputs for keys of type `InputT` to noise? Specially for tasks with many inputs. If the mocks is wrong you’ll have to potentially skim through many inputs in the error message to see if it’s missing or the mocked parameters are wrong.

acet: Hi @AndersonReyes thanks for the comment. Im afraid my understanding of flytekit is limited to my use of it, I'm not quite understanding what you mean here. Could you perhaps outline an example and I will try to create a test for it when I have a little time?

AndersonReyes: Actually ignore me you are already doing it. For some reason I thought fixedOutputs here was all the mocked inputs for the entire workflow (and that’s why I said maybe we need filter) but it’s not. fixedOutputs is just for this runnable node which makes more sense but ignore me in this comment.

andresgomezfrr (approved):
LGTM

PR #258, opened 2023-10-07
Bring back Blob support

honnix:
## _Read then delete this section_

_- Make sure to use a concise title for the pull-request._

_- Use #patch, #minor or #major in the pull-request title to bump the corresponding version. Otherwise, the patch version
will be bumped. [More details](https://github.com/marketplace/actions/github-tag-bump)_

# TL;DR
_Please replace this text with a description of what this PR accomplishes._

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added
 - [ ] Code documentation added
 - [ ] Any pending items have an associated Issue

## Complete description
 _How did you fix the bug, make the feature etc. Link to any design docs etc_

## Tracking Issue
_Remove the '*fixes*' keyword if there will be multiple PRs to fix the linked issue_

fixes https://github.com/flyteorg/flyte/issues/<number>

## Follow-up issue
_NA_
OR
_https://github.com/flyteorg/flyte/issues/<number>_

PR #259, opened 2023-10-08
Support struct for Java

honnix:
## _Read then delete this section_

# TL;DR

Support nested auto-value as input parameters.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
This is to add support for things like:
```java
@AutoValue
class Foo {
  public abstract SdkBindingData<String> foo();

  public abstract SdkBindingData<Bar> bar();
}

@AutoValue
class Bar {
  // note that this is not an SdkBindingData because it doesn't make sense since the whole class is wrapped as a promise
  // this also matches the direction of the existing but disabled IT
  public abstract String bar();
}
```

Scala layer is not supported yet because it is completely different mechanism.

## Tracking Issue
_NA

## Follow-up issue
_NA_


flytekit-jackson/src/main/java/org/flyte/flytekit/jackson/deserializers/StructDeserializer.java
```diff
@@ -29,23 +29,35 @@
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import org.flyte.api.v1.Literal;
-import org.flyte.api.v1.Scalar;
 import org.flyte.api.v1.Struct;
 import org.flyte.api.v1.Struct.Value;
+import org.flyte.flytekit.jackson.deserializers.StructDeserializer.StructWrapper;
 
-public class LiteralStructDeserializer extends StdDeserializer<Literal> {
```
honnix, line 37: It was confusing to use `Literal` as the target but expect a `Struct`.

flytekit-jackson/src/main/java/org/flyte/flytekit/jackson/serializers/GenericSerializer.java
```diff
@@ -48,85 +43,7 @@ public GenericSerializer(
...
-    gen.writeObject(kind);
-    gen.writeFieldName(STRUCT_VALUE);
-    writeTypeFunction.write(gen, structValue);
-    gen.writeEndObject();
-  }
-
-  interface WriteGenericFunction {
-    void write(JsonGenerator gen, Struct.Value value) throws IOException;
+    gen.writeFieldName(VALUE);
+    new StructSerializer().serialize(value.scalar().generic(), gen, serializerProvider);
```
honnix, line 47: The idea is to serialize a tree that can be deserialized into an auto-value object. which is exactly what `StructSerializer` does.

flytekit-jackson/src/main/java/org/flyte/flytekit/jackson/deserializers/LiteralMapDeserializers.java
```diff
@@ -1,41 +0,0 @@
...
-import com.fasterxml.jackson.databind.BeanDescription;
-import com.fasterxml.jackson.databind.DeserializationConfig;
-import com.fasterxml.jackson.databind.JavaType;
-import com.fasterxml.jackson.databind.JsonDeserializer;
-import com.fasterxml.jackson.databind.deser.Deserializers;
-import java.util.Map;
-import org.flyte.api.v1.LiteralType;
-import org.flyte.flytekit.jackson.JacksonLiteralMap;
-
-public class LiteralMapDeserializers extends Deserializers.Base {
```
honnix, line 28: This is no longer be used since we don't do a dynamic lookup anymore.

honnix:
I can no longer run IT locally because testcontainers generates a warning `Unable to mount a file from test host into a running container. This may be a misconfiguration or limitation of your Docker environment. Some features might not work.` so I cannot mount things properly during IT.

honnix:
> I can no longer run IT locally because testcontainers generates a warning `Unable to mount a file from test host into a running container. This may be a misconfiguration or limitation of your Docker environment. Some features might not work.` so I cannot mount things properly during IT.

Trying a fix in #261.

.scalafmt.conf
```diff
@@ -1,3 +1,3 @@
-version=2.5.2
```
honnix, line 1: This does not support `scala212source3`.

flytekit-jackson/src/test/java/org/flyte/flytekit/jackson/JacksonSdkTypeTest.java
```diff
@@ -574,49 +594,19 @@ public static AutoValueInput create(
...
         SdkBindingData<List<Map<String, String>>> lm,
         SdkBindingData<Map<String, List<String>>> ml,
         SdkBindingData<Map<String, Map<String, String>>> mm) {
       return new AutoValue_JacksonSdkTypeTest_AutoValueInput(
-          i, f, s, b, t, d, blob, l, lb, m, mb, ll, lm, ml, mm);
-    }
-  }
-
-  @AutoValue
-  public abstract static class StructInput {
```
honnix, line 591: I only discovered these classes in a very late stage so I did not bother adapting test cases to use these.

andresgomezfrr (approved):
LGTM 🚀

honnix:
This branch targets blob branch. Now #258 is a bit hard to review.

andresgomezfrr:
> This branch targets blob branch. Now #258 is a bit hard to review.

oh! I didn't notice that, don't worry, I will continue with the blob's PR

PR #262, opened 2023-10-11
Struct scala

honnix:
# TL;DR
On top of #259 to support Scala case class as struct.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
Basically to support this type of IO:
```scala
case class NestedNestedNested(string: String)
case class NestedNested(double: Double, nested: NestedNestedNested)
case class Nested(
    boolean: Boolean,
    byte: Byte,
    short: Short,
    int: Int,
    long: Long,
    float: Float,
    double: Double,
    string: String,
    list: List[String],
    map: Map[String, String],
    optBoolean: Option[Boolean],
    optByte: Option[Byte],
    optList: Option[List[String]],
    optMap: Option[Map[String, String]],
    nested: NestedNested
)

case class NestedIOTaskInput(
    @Description("the name of the person to be greeted")
    name: SdkBindingData[String],

    @Description("a nested input")
    generic: SdkBindingData[Nested]
)
```

## Tracking Issue
_NA_

## Follow-up issue
_NA_

### DIFF
--- a/flytekit-examples-scala/src/main/scala/org/flyte/examples/flytekitscala/NestedIOTask.scala
+++ b/flytekit-examples-scala/src/main/scala/org/flyte/examples/flytekitscala/NestedIOTask.scala
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2023 Flyte Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.flyte.examples.flytekitscala
+
+import org.flyte.flytekit.{SdkBindingData, SdkRunnableTask, SdkTransform}
+import org.flyte.flytekitscala.{
+  Description,
+  SdkBindingDataFactory,
+  SdkScalaType
+}
+
+case class NestedNestedNested(string: String)
+case class NestedNested(double: Double, nested: NestedNestedNested)
+case class Nested(
+    boolean: Boolean,
+    byte: Byte,
+    short: Short,
+    int: Int,
+    long: Long,
+    float: Float,
+    double: Double,
+    string: String,
+    list: List[String],
+    listOfNested: List[NestedNested],
+    map: Map[String, String],
+    mapOfNested: Map[String, NestedNested],
+    optBoolean: Option[Boolean],
+    optByte: Option[Byte],
+    optList: Option[List[String]],
+    optMap: Option[Map[String, String]],
+    nested: NestedNested
+)
+case class NestedIOTaskInput(
+    @Description("the name of the person to be greeted")
+    name: SdkBindingData[String],
+    @Description("a nested input")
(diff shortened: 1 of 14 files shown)


flytekit-examples-scala/src/main/scala/org/flyte/examples/flytekitscala/NestedIOTask.scala
```diff
@@ -0,0 +1,85 @@
...
+    list: List[String],
+    listOfNested: List[NestedNested],
+    map: Map[String, String],
+    mapOfNested: Map[String, NestedNested],
+    optBoolean: Option[Boolean],
+    optByte: Option[Byte],
+    optList: Option[List[String]],
+    optMap: Option[Map[String, String]],
+    nested: NestedNested
+)
```
andresgomezfrr, line 46: I'm wondering now... what happens if you try to build a nested input using outputs from another task? I guess this is not possible because the outputs will be `SdkBindingData[_]` and the nested input uses plain types, right?

flytekit-scala_2.13/src/main/scala/org/flyte/flytekitscala/SdkScalaType.scala
```diff
@@ -231,6 +233,15 @@ object SdkScalaType {
   implicit def durationLiteralType: SdkScalaLiteralType[Duration] =
     DelegateLiteralType(SdkLiteralTypes.durations())
 
+  // more specific matching to fail the usage of SdkBindingData[Option[_]]
+  implicit def optionLiteralType: SdkScalaLiteralType[Option[_]] = ???
```
andresgomezfrr, line 237: I think if we support this in the future, it will be well received by scala users.

flytekit-examples-scala/src/main/scala/org/flyte/examples/flytekitscala/NestedIOTask.scala
honnix: This was discussed briefly internally before, and I don't think there is any valid use case could make it troublesome.

Assuming output is `foo = SdkBindingData[Foo]`, to use the value in a nested structure, it would be just `Bar(foo=foo)`, where `Bar` is `class Bar(foo: SdkBindingData[Foo])`. They key point is: nested structure is always plain type, no matter input or output, so they should always match.

flytekit-scala_2.13/src/main/scala/org/flyte/flytekitscala/SdkScalaType.scala
honnix: I'm not sure whether it makes sense to support `null` value for protobuf message. There is no real `null` in protobuf, and protobuf struct supports null value in a special way. If user passes in `None`, we will use default value for serialization, and when the task receives that, it needs to understand whether the input was actually `None`, or was it really an empty string for example.

I have not thought about this in depth, but right now I don't see much value of it.

flytekit-examples-scala/src/main/scala/org/flyte/examples/flytekitscala/NestedIOTask.scala
honnix: Note that this is not something special to Scala and it is the same for Java API.

andresgomezfrr (approved):

flytekit-scala_2.13/src/main/scala/org/flyte/flytekitscala/SdkScalaType.scala
honnix: Yeah. I hope they normally do not need to use "null value" for signalling because it is not a good practice using protobuf.

PR #263, opened 2023-10-18
Case class for BindingCollection/Map

honnix:
# TL;DR
Make Scala `BindingCollection` and `BindingMap` case classes.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
Making them case classes to be able to support equality check. They should have been case classes in the first place, so this is a bug fix.

This change should have no user impact because both classes are package private.

## Tracking Issue
_NA_

## Follow-up issue
_NA_

### DIFF
--- a/flytekit-scala_2.13/src/main/scala/org/flyte/flytekit/BindingCollection.scala
+++ b/flytekit-scala_2.13/src/main/scala/org/flyte/flytekit/BindingCollection.scala
@@ -23,7 +23,7 @@ import org.flyte.flytekitscala.SdkLiteralTypes.collections
 import java.util.function
 import scala.collection.JavaConverters._
 
-private[flyte] class BindingCollection[T](
+private[flyte] case class BindingCollection[T](
     elementType: SdkLiteralType[T],
     bindingCollection: List[SdkBindingData[T]]
 ) extends SdkBindingData[List[T]] {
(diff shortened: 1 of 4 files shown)


flytekit-scala_2.13/src/main/scala/org/flyte/flytekit/BindingCollection.scala
```diff
@@ -17,13 +17,12 @@
 package org.flyte.flytekit
```
honnix, line 17: BTW why these classes are not under `flytekitscala` package?

honnix:
```
We cannot make it protected either as it would
  * be good for the own object but both implementations deal with list or maps
  * of [[SdkBindingData]] and therefore cannot call this method because it is in
  * a different class.
```
Not sure I understand what this refers to.

honnix: I think I understand what this means.

honnix: I tried revising it a bit in a latter commit.

andresgomezfrr (approved):

PR #265, opened 2023-10-27
Minor clean up of struct and blob

honnix:
# TL;DR
Minor clean up of struct and blob

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
Mostly getting unit tests in place. I didn't do this in previous PR because we did not realize the test cases are in a dedicated maven module.

## Tracking Issue
_NA_
## Follow-up issue
_NA_

### DIFF
--- a/flytekit-scala_2.13/src/main/scala/org/flyte/flytekitscala/SdkBindingDataConverters.scala
+++ b/flytekit-scala_2.13/src/main/scala/org/flyte/flytekitscala/SdkBindingDataConverters.scala
@@ -150,7 +150,10 @@ object SdkBindingDataConverters {
               SdkScalaLiteralTypes.strings(),
               jf.Function.identity()
             )
-          case SimpleType.STRUCT => ??? // TODO not yet supported
+          case SimpleType.STRUCT =>
+            throw new UnsupportedOperationException(
+              "Converting Scala case class instance to Java object is not supported"
+            )
           case SimpleType.BOOLEAN =>
             TypeCastingResult(
               SdkScalaLiteralTypes.booleans(),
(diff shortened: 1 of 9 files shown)


flytekit-scala_2.13/src/main/scala/org/flyte/flytekitscala/SdkLiteralTypes.scala
```diff
@@ -66,6 +66,10 @@ object SdkLiteralTypes {
         datetimes().asInstanceOf[SdkLiteralType[T]]
       case t if t =:= typeOf[Duration] =>
         durations().asInstanceOf[SdkLiteralType[T]]
+      case t if t =:= typeOf[Blob] =>
+        blobs(BlobType.DEFAULT).asInstanceOf[SdkLiteralType[T]]
+      case t if t <:< typeOf[Product] && !(t =:= typeOf[Option[_]]) =>
```
honnix, line 71: Kinda best effort matching, not at all precise. I'm wondering whether this method has to be public as it is not used by anything but test case.

flytekit-scala_2.13/src/main/scala/org/flyte/flytekitscala/SdkBindingDataConverters.scala
```diff
@@ -150,7 +150,10 @@ object SdkBindingDataConverters {
               SdkScalaLiteralTypes.strings(),
               jf.Function.identity()
             )
-          case SimpleType.STRUCT => ??? // TODO not yet supported
+          case SimpleType.STRUCT =>
```
honnix, line 153: Be explicit that we do not support this intentionally.

Support Scala case class instance to auto-value object is not trivial and it would introduce dependency on flytekit-jackson in one way or another. Hopefully in practice we don't really have need of this.

andresgomezfrr (approved):

PR #266, opened 2023-11-22
Handle Option correctly

honnix:
# TL;DR
Handle Scala `Option` correctly by using correct type.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
For case classes like:
```scala
case class ScalarNested(
    foo: String,
    bar: Option[String],
    nestedNested: Option[ScalarNestedNested]
)
case class ScalarNestedNested(foo: String, bar: Option[String])
```
the current code doesn't handle typing correctly because calling `param.typeSignature.dealias.typeArgs.head.typeSymbol` on `Option` type does not give what we need, and it should be `param.typeSignature.dealias.typeArgs.head` instead.

Also we are not handling nested `Option` correctly, although in practice there is no good reason to define `Option[Option[String]]`.

## Tracking Issue
_NA_

## Follow-up issue
_NA_

### DIFF
--- a/flytekit-scala_2.13/src/main/scala/org/flyte/flytekitscala/SdkLiteralTypes.scala
+++ b/flytekit-scala_2.13/src/main/scala/org/flyte/flytekitscala/SdkLiteralTypes.scala
@@ -297,41 +297,35 @@ object SdkLiteralTypes {
     ): S = {
       val mirror = runtimeMirror(classTag[S].runtimeClass.getClassLoader)
 
-      def valueToParamValue(value: Any, param: Symbol): Any = {
-        def valueToParamValue0(value: Any, param: Symbol): Any = {
-          if (param.typeSignature =:= typeOf[Byte]) {
-            value.asInstanceOf[Double].toByte
-          } else if (param.typeSignature =:= typeOf[Short]) {
-            value.asInstanceOf[Double].toShort
-          } else if (param.typeSignature =:= typeOf[Int]) {
-            value.asInstanceOf[Double].toInt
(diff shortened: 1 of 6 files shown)


flytekit-scala_2.13/src/main/scala/org/flyte/flytekitscala/SdkLiteralTypes.scala
```diff
@@ -297,41 +297,39 @@ object SdkLiteralTypes {
...
+          value.asInstanceOf[Double].toShort
+        } else if (tpe =:= typeOf[Int]) {
+          value.asInstanceOf[Double].toInt
+        } else if (tpe =:= typeOf[Long]) {
+          value.asInstanceOf[Double].toLong
+        } else if (tpe =:= typeOf[Float]) {
+          value.asInstanceOf[Double].toFloat
+        } else if (tpe <:< typeOf[Option[Any]]) { // this has to be before Product check because Option is a Product
+          if (value == None) { // None is used to represent Struct.Value.Kind.NULL_VALUE when converting struct to map
+            None
```
honnix, line 313: This is another bug fix. We have been converting things into `Some(None)`, due to type conformation test failure against `T` in `Option[T]`, which is exactly what this PR set off to fix. Bug on top of bug :(

Hopefully the extended unit tests and ITs cover better.

andresgomezfrr (approved):

PR #267, opened 2023-11-23
Fix nested list & map of structs using case classes

andresgomezfrr:
# TL;DR
Fix nested list of structs using scala layer.

```scala
case class ScalarNested(
    foo: String,
    bar: Option[String],
    nestedNested: Option[ScalarNestedNested],
    nestedNestedList: List[ScalarNestedNested], // <----- THIS
    nestedNestedMap: Map[String, ScalarNestedNested] // <----- THIS
)
case class ScalarNestedNested(foo: String, bar: Option[String])
```

support-nested-list-of-structs

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

### DIFF
--- a/flytekit-scala-tests/src/test/scala/org/flyte/flytekitscala/SdkScalaTypeTest.scala
+++ b/flytekit-scala-tests/src/test/scala/org/flyte/flytekitscala/SdkScalaTypeTest.scala
@@ -52,7 +52,9 @@ import org.flyte.flytekitscala.SdkLiteralTypes.{
 case class ScalarNested(
     foo: String,
     bar: Option[String],
-    nestedNested: Option[ScalarNestedNested]
+    nestedNested: Option[ScalarNestedNested],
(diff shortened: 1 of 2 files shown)


flytekit-scala-tests/src/test/scala/org/flyte/flytekitscala/SdkScalaTypeTest.scala
```diff
@@ -191,6 +192,18 @@ class SdkScalaTypeTest {
                     "bar" -> Struct.Value.ofStringValue("bar")
                   ).asJava
                 )
+              ),
+              "nestedList" -> Struct.Value.ofListValue(
```
honnix, line 196:
```suggestion
              "nestedNestedList" -> Struct.Value.ofListValue(
```

honnix (approved):

PR #272, opened 2023-12-11
Make plugin task first-class citizen

honnix:
This is the same as #268, which was not ready to get merged and got reverted later.

# TL;DR
Create a dedicated Plugin Task instead of abusing Runnable Task.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
* ~~Manually pick proto change from unmerged PR https://github.com/flyteorg/flyte/pull/4107~~ We can go back to this later, for now I have removed this part to make this PR shippable
* Now we have a proper Plugin Task so users will no longer to have implement Runnable Task with an a `run` returning `null`

Note that it is still supported to use Runnable Task to implement tasks that are handled by backend plugin, for backward compatibility.

## Tracking Issue
_NA_

## Follow-up issue
_NA_

### DIFF
--- a/flyteidl-protos/src/main/proto/flyteidl/core/tasks.proto
+++ b/flyteidl-protos/src/main/proto/flyteidl/core/tasks.proto
@@ -59,14 +59,6 @@ message RuntimeMetadata {
 
     //+optional It can be used to provide extra information about the runtime (e.g. python, golang... etc.).
     string flavor = 3;
-
-    //+optional It can be used to provide extra information for the plugin.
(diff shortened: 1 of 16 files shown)


honnix:
I dropped the `isSyncPlugin` field from this PR because the upstream might take some time.

andresgomezfrr (approved):

PR #283, opened 2024-02-02
Escape $ inside the task name

andresgomezfrr:
# TL;DR
Escape `$` from the task name. Normally task name is generated from the className and if the className is inside and scala object the result is `ObjectName$TaskName`, so we need to scape the `$` to can execute the tasks properly.

Tested:
Screenshot 2024-02-02 at 12 55 17

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

### DIFF
--- a/jflyte-utils/src/main/java/org/flyte/jflyte/utils/ProjectClosure.java
+++ b/jflyte-utils/src/main/java/org/flyte/jflyte/utils/ProjectClosure.java
@@ -28,6 +28,8 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
+import com.google.common.escape.Escaper;
+import com.google.common.escape.Escapers;
 import com.google.protobuf.ByteString;
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


jflyte-utils/src/main/java/org/flyte/jflyte/utils/ProjectClosure.java
```diff
@@ -92,6 +94,8 @@ public abstract class ProjectClosure {
 
   public abstract Map<LaunchPlanIdentifier, LaunchPlan> launchPlans();
 
+  private static final Escaper escaper = Escapers.builder().addEscape('$', "\\$").build();
```
honnix, line 97:
```suggestion
  private static final Escaper ESCAPER = Escapers.builder().addEscape('$', "\\$").build();
```

jflyte-utils/src/main/java/org/flyte/jflyte/utils/ProjectClosure.java
```diff
@@ -483,7 +487,7 @@ static TaskTemplate createTaskTemplateForRunnableTask(RunnableTask task, String
                     "jflyte",
                     "execute",
                     "--task",
-                    task.getName(),
+                    escaper.escape(task.getName()),
```
honnix, line 490:
```suggestion
                    ESCAPER.escape(task.getName()),
```

honnix (approved):

PR #308, opened 2024-07-04
#patch Adding support for SdkBindingData[scala.Option[_]]

jschuchart-spot:
# TL;DR
Adding explicit support for scala.Option type

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added
 - [x] Code documentation added

## Complete description
Added explicit special case for `Option` type when instantiating a `generics` literal type. Since support for any case classes in SdkLiteralTypes has been introduced, it's not immediately obvious why this works
```
case class Example(value: Option[String])
SdkLiteralTypes.of[Example]()
```
but this doesn't:
```
SdkLiteralTypes.of[Option[String]]()
```

The latter case is especially useful for input classes to workflows and SdkTransforms of the form
```
case class Input(
  inputStr: SdkBindingData[String],
  optional: SdkBindingData[Option[String]],
  ...
)
```

The Option type can't be instantiated directly like a common case class, so it needs to be instantiated differently.
I believe the common usage of Option as a field type in scala classes justifies this special handling.

## Tracking Issue
_NA_

## Follow-up issue
_NA_

### DIFF
--- a/flytekit-scala_2.13/src/main/scala/org/flyte/flytekitscala/SdkLiteralTypes.scala
+++ b/flytekit-scala_2.13/src/main/scala/org/flyte/flytekitscala/SdkLiteralTypes.scala
@@ -28,6 +28,7 @@ import scala.reflect.api.{Mirror, TypeCreator, Universe}
 import scala.reflect.runtime.universe
 import scala.reflect.{ClassTag, classTag}
 import scala.reflect.runtime.universe.{
+  ClassSymbol,
   NoPrefix,
   Symbol,
   Type,
@@ -72,7 +73,7 @@ object SdkLiteralTypes {
         blobs(BlobType.DEFAULT).asInstanceOf[SdkLiteralType[T]]
       case t if t =:= typeOf[Binary] =>
         binary().asInstanceOf[SdkLiteralType[T]]
-      case t if t <:< typeOf[Product] && !(t =:= typeOf[Option[_]]) =>
+      case t if t <:< typeOf[Product] =>
         generics().asInstanceOf[SdkLiteralType[T]]
 
       case t if t =:= typeOf[List[Long]] =>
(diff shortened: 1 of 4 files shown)


flytekit-scala_2.13/src/main/scala/org/flyte/flytekitscala/SdkScalaType.scala
```diff
@@ -232,11 +232,8 @@ object SdkScalaType {
   implicit def durationLiteralType: SdkScalaLiteralType[Duration] =
     DelegateLiteralType(SdkLiteralTypes.durations())
 
-  // more specific matching to fail the usage of SdkBindingData[Option[_]]
-  implicit def optionLiteralType: SdkScalaLiteralType[Option[_]] = ???
-
   // fixme: using Product is just an approximation for case class because Product
-  // is also super class of, for example, Option and Tuple
+  // is also super class of, for example, Either or Try
```
jschuchart-spot, line 236: Tuple is actually also just a case class and already works with the current implementation. The problem arises when there is an abstract Product type with specific implementations like `Either` or `Try`

flytekit-scala_2.13/src/main/scala/org/flyte/flytekitscala/SdkLiteralTypes.scala
```diff
@@ -391,24 +392,37 @@ object SdkLiteralTypes {
...
+      }
+
+      val clazz = typeOf[S].typeSymbol.asClass
+      // special handling of scala.Option as it is a Product, but can't be instantiated like common
+      // case classes
+      if (clazz.name.toString == "Option")
+        map
+          .get("value")
+          .map(valueToParamValue(_, typeOf[S].typeArgs.head))
+          .asInstanceOf[S]
```
jschuchart-spot, line 422: Since `map.get(...)` already produces an Option, we can just go with that and map the inner value if present

flytekit-scala_2.13/src/main/scala/org/flyte/flytekitscala/SdkLiteralTypes.scala
```diff
@@ -391,24 +392,37 @@ object SdkLiteralTypes {
...
+      else
+        instantiateViaConstructor(clazz)
```
jschuchart-spot, line 424: pulled out for readability

flytekit-scala_2.13/src/main/scala/org/flyte/flytekitscala/SdkLiteralTypes.scala
```diff
@@ -72,7 +73,7 @@ object SdkLiteralTypes {
         blobs(BlobType.DEFAULT).asInstanceOf[SdkLiteralType[T]]
       case t if t =:= typeOf[Binary] =>
         binary().asInstanceOf[SdkLiteralType[T]]
-      case t if t <:< typeOf[Product] && !(t =:= typeOf[Option[_]]) =>
+      case t if t <:< typeOf[Product] =>
```
jschuchart-spot, line 76: This check for avoiding Option was not successful to begin with. It would need to use `<:<` instead of `=:=`.

flytekit-scala_2.13/src/main/scala/org/flyte/flytekitscala/package.scala
```diff
@@ -30,7 +30,11 @@ package object flytekitscala {
     } catch {
       case _: Throwable =>
         // fall back to java's way, less reliable and with limitations
-        product.getClass.getDeclaredFields.map(_.getName).toList
+        val methodNames = product.getClass.getDeclaredMethods.map(_.getName)
+        product.getClass.getDeclaredFields
+          .map(_.getName)
+          .filter(methodNames.contains)
+          .toList
```
jschuchart-spot, line 37: This gave me `["serialVersionUUID", "value"]` for `Option` which meant that the stored map used "serialVersionUUID" as the key for the value because this list is zipped with the `productElements`.

andresgomezfrr (commented):
What do you think about add a specific `of` factor for Options.

`ofOption(value) => SdkBindingData[Option[_]]`

and 

`of(Option(value)) => SdkBindingData[Option[_]]`

to avoid to need to use the generics

jschuchart-spot:
> What do you think about add a specific `of` factor for Options.
> 
> `ofOption(value) => SdkBindingData[Option[_]]`
> 
> and
> 
> `of(Option(value)) => SdkBindingData[Option[_]]`
> 
> to avoid to need to use the generics

How and why do you mean to avoid generics exactly? You'd still want a `SdkBindingData[Option[T]]` as a result like you have for collections and maps. To avoid generics entirely would require adding `ofStringOption`, `ofIntegerOption`, etc. which may be bloating the code.
What could be useful in my opinion may be `ofOption[T](value: T): SdkBindingData[Option[T]]` and/or `of[T : TypeTag](option: Option[T]): SdkBindingData[Option[T]]`. The reason being that this `of` factory will otherwise only work if it is a defined Option (similar to collections and maps needing to contain at least one value), which seems an odd restriction in this case.

andresgomezfrr (approved):

